PR #1, opened 2021-03-18
First preliminary attempt at normalisation

MatthewDaggitt:
At the moment, it completely ignores deBruijn indices so not proposing to merge in at the moment, more for feedback. First time ever writing a piece of Haskell, or indeed any piece of code with non-trivial control in a functional language, so all and any feedback however minor is very welcome! 

My current plan on that front is to first write a module that takes the AST with users names and converts it into an AST into deBruign indices, and only then come back and try and get deBruign substitution working. Sound like a plan @wenkokke?

### DIFF
--- a/src/hs/Vehicle/Core/Normalise.hs
+++ b/src/hs/Vehicle/Core/Normalise.hs
@@ -13,7 +13,12 @@ module Vehicle.Core.Normalise
   ) where
 
 import Vehicle.Core.Type ( Sort (..), Expr (..), Sort, EArg (..) )
+<<<<<<< HEAD
 import Vehicle.Core.Instance.Recursive (ExprF (..))
+=======
+import Vehicle.Core.DeBruijn (DeBruijn, SortedDeBruijn)
+import Vehicle.Core.Type.Instance.Recursive (ExprF (..))
+>>>>>>> First preliminary attempt at normalisation
 import Control.Monad.Except (MonadError, Except, runExcept)
 import Data.Map (Map, lookup, insert, notMember)
 import Data.Functor.Foldable (fold)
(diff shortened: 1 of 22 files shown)


src/hs/Vehicle/Core/Normalise.hs
```diff
@@ -0,0 +1,109 @@
...
+
+normApp :: (MonadNorm m) => Map DeBruijn (NormExpr ann) -> 
+            m (NormExpr ann) -> m (NormExpr ann) -> m (NormExpr ann) 
+normApp ctxt mFn mArg = do
+    fn <- mFn
+    arg <- mArg
+    varName <- lamVarName fn
+    varBody <- lamBody fn
+    updatedCtxt <- updateCtxt ctxt varName arg
+    norm updatedCtxt varBody
```
MatthewDaggitt: Bit worried about this call and the similar one in `normLet`. If I understand the semantics of the `Vehicle.Core.Instance.Recursive` the fold operator is already normalising all the sub-expressions for us in each case. That means that this is the second time the body is being normalised. The problem is of course that it's only in this case that we have the function's bound variable in the context....

Does that mean I can't use the recursive instance available in `Vehicle.Core.Instance.Recursive`?

src/hs/Vehicle/Core/Normalise.hs
```diff
@@ -0,0 +1,109 @@
...
+  norm :: MonadNorm m => Map DeBruijn vf -> vf -> m vf
+
+instance Norm (NormExpr ann) where
+  norm ctxt = fold $ \case
+    EVarF ann (SortedDeBruijn i) -> normVar ctxt ann i
+    ELetF _ arg exp1 exp2 -> normLet ctxt (index arg) exp1 exp2
+    EAppF _ exp1 exp2 -> normApp ctxt exp1 exp2
+
+    ELitSeqF ann exprs -> ELitSeq ann <$> sequence exprs
+    EAnnF ann expr typ -> EAnn ann <$> expr <*> pure typ
```
MatthewDaggitt: Is there a function that does `\v -> _<*> pure v` automatically? Couldn't find one...

MatthewDaggitt:
Added in 565e86afcec9c0c810ff09907084223f319837f0

PR #8, opened 2021-07-08
Make scope and type checking work

bobatkey:
Fix several bugs that prevented scope and type checking from working:
1. Hanging bug in scope checking, caused by circular instance definitions in `DataflowT`
2. Scope checking properly handles toplevel definitions
3. Type checking doesn't hang, which was caused by circular `Eq` definition in `Info`
4. Type equality checking ignores provenance information (though this is a bit of a hack: the `Eq` instance for `Provenance` makes all provenance equal)
5. Added some Show instances for debugging
6. Made the Agda output go to stdout if there is no output file specified


MatthewDaggitt:
Thank you so much @bobatkey, looks amazing!

PR #10, opened 2021-07-13
Fix de Bruijn -> Symbol conversion; fixes #9

### LINKED ISSUE #9: de Bruijn -> named needs to process annotations at the same time as the rest of the AST
This issue arises from trying to get the following to type check:
```
polyId : forall t. t -> t
polyId = \{t} -> \x -> x
```
Vehicle currently crashes with `Error: DeBruijn index 0 greater than current context size 0 (Line 3, Columns 20-21)` (perhaps with different line numbers). This error is thrown by the conversion from de Bruijn back to concrete named representation.

Currently the de Bruijn to concrete names conversion works in two stages:
1. de Bruijn indices in the AST are converted to `Name`s, which either copies over the user's original name, or uses `Machine` for machine generated binders
2. Then a second pass converts all `Name`s to `Symbol`s (so they can be printed) and *simultaneously* converts all the annotations (which contain the types of each sub term) from de Bruijn to names (and then symbols).

The problem is that when the second pass converts annotations from de Bruijn to Names, it operates on each annotation in isolation without taking into account the binding context. The variable occurrence `x` at the end is annotated by the type checker with the type `TVar 0`, where the de Bruijn index `0` makes sense in the content of the binder `{t}`. However, the conversion of the annotation `TVar 0` does not know about this context yielding the error above.

The solution therefore is to convert de Bruijn to `Name`s in the main AST and the annotations in the same past.

However, this violates the current setup of the `DataflowT` monad transformer. Processing of type and expression arguments (sorts `'TARG` `'EARG`) is assumed to only result in bindings being added to the context. But arguments are annotated with their kind/type, so they need converting from de Bruijn to `Name`s. To do this, processing needs access to the current context which is not allowed by the current definition of `DataflowT`.

Two solutions come to mind:
1. Do not treat arguments as a special kind, so they get handled by their surrounding AST node.
2. Tweak `DataflowT` to be more permissive, and use a state monad for arguments.

bobatkey:
Fixes the conversion from deBruijn to Symbols, via Names. The original problem was described in issue #9. The fix is to convert the main AST and the ASTs in the annotations in a single pass. This required modifying the `DataflowT` monad to treat allow for context information to always flow down the tree (specifically, into type and expression arguments).

This update makes the polymorphic identity examples in `examples/polyI.vcl` type check. However, there seems to still be some bug that prevents polymorphic functions with multiple arguments from working -- see the file `examples/polyK.vcl`. I suspect that there is something wrong with the substitution implementation.

### DIFF
--- a/examples/polyI.vcl
+++ b/examples/polyI.vcl
@@ -1,0 +1,17 @@
+polyId : forall t. t -> t
+polyId = \{t} -> \x -> x
+
+realId : Real -> Real
+realId = polyId {Real}
+
+intId : Int -> Int
+intId = polyId {Int}
+
+real : Real
+real = polyId {Real} 0.0
+
+int : Int
+int = polyId {Int} 0
+
+tensorId : forall t. Tensor t [2] -> Tensor t [2]
+tensorId {t} = polyId {Tensor t [2]}
--- a/examples/polyK.vcl
+++ b/examples/polyK.vcl
@@ -1,0 +1,8 @@
+k : forall s. forall t. s -> t -> s
+k {s} {t} x y = x
+
+k2 : forall b. forall a. a -> b -> a
+k2 {b} {a} = k {a} {b}
+
+kRealInt : Real -> Int -> Real
+kRealInt = k {Real} {Int}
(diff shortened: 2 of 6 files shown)


bobatkey:
The second commit fixes the `examples/polyK.vcl`. Type variables were being added in the wrong order. oops!

PR #45, opened 2021-11-16
added installation instructions

Yiergot:
initial guide to get started with vehicle, won't be final README

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,6 @@
    - bnfc >2.9
        - `stack update`
        - `stack install BNFC`
-   - if ubuntu
-       - add `allow-newer: true` to stack.yaml file (not sure if this is the way to go please correct me)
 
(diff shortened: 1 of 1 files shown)


wenkokke:
Not sure this is exactly the right way, and I’d like to stay away from `—allow-newer` as it is *bonkers*.

wenkokke:
I think the right way to go is to call `stack install bnfc` from the vehicle directory.
We could do from the Makefile if BNFC isn’t found, and let `BNFC=stack exec bnfc`?

Yiergot:
just tested on the vm, 'stack update' and then 'stack install BNFC' (case sensitive) works as well. Can remove the thing about binary download.

wenkokke:
@MatthewDaggitt and I are working on a more comprehensive fix

PR #50, opened 2021-11-23
Try to fix the CI

MatthewDaggitt:
Made this a PR to avoid making endless minor commits.

### DIFF
--- a/.github/workflows/vehicle-tests.yml
+++ b/.github/workflows/vehicle-tests.yml
@@ -36,12 +36,19 @@ jobs:
             ~/.cabal
           key: ${{ matrix.os }}-ghc-${{ matrix.ghc }}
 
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
Okay, well we're installing `bnfc`/`alex`/`happy` but the `findExecutable` and `cmd` commands can't find them at all... Giving up on this for today.

wenkokke:
Once we have this working shell we add operating systems to the testing matrix as well? Makes sure we don't lose that windows compatibility?

MatthewDaggitt:
Absolutely. Just need to get it working first :smile:

wenkokke:
Let me know if you would like some help with this. I have spent more hours than I am willing to admit wrestling with a CI.

Issue #97, opened 2022-04-21
Allow user to specify which function generates loss function & normalisation


ndslusarz:
--functionName 

We need to stop full normalisation. Will require some normalisation (declaration substitution has to happen inside the function but we do need the name of the one generating loss)

Annotations - not meaningfull, get rid of in translation
Let - we want them normalised out (could keep them in but performance gain not worth added complexity)
Lam - can't be normalised, needs to be translated and put into JSON

Important: Loss functions can ONLY be generated if the function goes to Prop/Bool. Otherwise we need to throw an error.

MatthewDaggitt:
Hmm I think all these things have been done so now closing.

PR #116, opened 2022-05-23
Loss functions

### LINKED ISSUE #96: More meaningful names in the generated json

ndslusarz:
Has work for issues #96 (changing the names to more meaningful - finished) and #97.

 #97 is partially solved - there is code handling Annotations, Lam and Let which we don't want normalised and checks type of function making sure it goes to Bool/Prop (but it's still missing the new normalisation to replace the full normalisation we were using before).


src/hs/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -35,17 +36,18 @@ instance ToJSON Domain where
...
 data LExpr
-  = Neg LExpr
-  | Con Double
+  = Negation LExpr
+  | Constant Double
   | Min LExpr LExpr
   | Max LExpr LExpr
-  | Sub LExpr LExpr
-  | Ind LExpr LExpr
+  | Substraction LExpr LExpr
```
MatthewDaggitt: Subtraction?

src/hs/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -35,17 +36,18 @@ instance ToJSON Domain where
...
+  | IndicatorFunction LExpr LExpr
   | Var V.DBIndex
```
MatthewDaggitt: As we're being verbose, can we have `Variable`, `NetworkApplication`, `Quantifier` and `TensorLiteral` as well?

MatthewDaggitt (changes requested):
Otherwise looks good!

src/hs/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Corrected now :)

src/hs/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Done in the new commit.

MatthewDaggitt (approved):

MatthewDaggitt:
Fixed the merge conflicts, but when I run the tests via `cabal run build test`, I get an error saying there's still no case for translating variables! We'd probably best add one!

PR #117, opened 2022-05-27
first attempt at python loss function generation

Tgl70:


src/hs/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -46,6 +46,7 @@ data LExpr
   | Quant Quantifier Symbol Domain LExpr
   | At LExpr LExpr
   | TensorLit [LExpr]
+  | Lambda Symbol LExpr
```
MatthewDaggitt, line 58: Could you remove this change from this PR? I think Natalia has already added it in hers.

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+
+    jsonDICT = generateJson(pathToSpec, functionName)
+    context = []
+    loss = '_'
+    loss = translateJSONtoLOSSfunction(resources, jsonDICT)
+
+    return loss(context)
+    
+
+def generateJson(pathToSpec, functionName):
```
MatthewDaggitt: Please could we use `snake_case` consistently everywhere throughout the files?

And also add type annotations to all arguments, and the return types of functions?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+    #call vehicle executable on the provided spec file - for now just load a specific json file
+    pathToJSON = '/Users/marcocasadio/Projects/vehicle/test/Test/Compile/Golden/bounded/bounded-output.json'
+    f = open(pathToJSON)
+    jsonDICT = json.load(f)
+    f.close()
```
MatthewDaggitt: This is unsafe as if `json.load` throws an exception then the file with never be closed. We should use `with` to open resources (e.g. https://www.pythonforbeginners.com/files/with-statement-in-python)

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+    return jsonDICT
+
+
+def translateJSONtoLOSSfunction(resources, jsonDICT):
+    if isinstance(jsonDICT, dict):
```
MatthewDaggitt: Always try to return early. Much better to say 
```
if not isinstance (jsonDICT, dict)
  return

....
``` 

Although this test shouldn't even be necessary if we add a type annotation to `jsonDICT`?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+                return loss1(context) - loss2(context)
+
+            return result_func
+
+        elif tag == 'Ind':
+            loss1 = translateJSONtoLOSSfunction(resources, contents[0])
+            loss2 = translateJSONtoLOSSfunction(resources, contents[1])
+
+            def result_func(context):
+                if loss1(context) == loss2(context):
```
MatthewDaggitt: Much neater to write this as `return 1 if loss1(context) == loss2(context) else 0`?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+            if contents[0] == 'All':
+                def result_func(context):
+                    domain = domainloss(context)
+                    maxLoss = np.NINF
+                    for _ in range(10):
+                        context.insert(0, random.uniform(domain[0], domain[1]))
+                        value = bodyloss(context)
+                        if value > maxLoss:
+                            maxLoss = value
+                        context.pop(0)
```
MatthewDaggitt: I'm unsure what the efficiency of adding and removing elements at the start of a Python list is? Regardless, we should probably be explicitly using a stack data structure?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+def generateJson(pathToSpec, functionName):
+    #call vehicle executable on the provided spec file - for now just load a specific json file
+    pathToJSON = '/Users/marcocasadio/Projects/vehicle/test/Test/Compile/Golden/bounded/bounded-output.json'
+    f = open(pathToJSON)
+    jsonDICT = json.load(f)
+    f.close()
+    return jsonDICT
+
+
+def translateJSONtoLOSSfunction(resources, jsonDICT):
```
MatthewDaggitt: We could just call this `to_loss_function` which is shorter, but still meaningful I think?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+            def result_func(context):
+                return lossTensor(context)[lossIndex(context)]
+
+            return result_func
+            
+        elif tag == 'NetApp':      
+            model = resources[contents[0]]
+            lossInputsList = []
+            for c in contents[1]:
+                lossInputsList.append(translateJSONtoLOSSfunction(resources, c))
```
MatthewDaggitt: This can be written as list comprehension, e.g.
```python
input_losses = [to_loss_function(resources, c) for c in contents [1]]
```

Indeed, several other cases in this file where a similar refactoring can be done, i.e. anywhere with the pattern:
```python
l = []
for c in ...
  l.append(..)

```

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+
+            if contents[0] == 'All':
+                def result_func(context):
+                    domain = domainloss(context)
+                    maxLoss = np.NINF
+                    for _ in range(10):
+                        context.insert(0, random.uniform(domain[0], domain[1]))
+                        value = bodyloss(context)
+                        if value > maxLoss:
+                            maxLoss = value
```
MatthewDaggitt: These three lines could be replaced with, `maxLoss = max(maxLoss, bodyloss(context))`?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+                        context.pop(0)
+                    return maxLoss
+
+            elif contents[0] == 'Any':
+                def result_func(context):
```
MatthewDaggitt: This function is identical to the one above, apart from using `min` rather than `max`. Maybe simply have one function and pass in either the `min` or `max` operation as an argument?

test/Test/Compile/Golden/bounded/bounded-output.json
```diff
@@ -1,130 +1,149 @@
-{
-  "tag": "Quant",
-  "contents": [
-    "All",
-    "x",
-    [],
```
MatthewDaggitt: Unfortunately this file is part of the test suite, so changing it will cause tests to fail. Could you create your own local copy of this file until we get around to updating the JSON generation to actually produce these changes?

test/specs/bounded.vcl
```diff
@@ -1,4 +1,4 @@
-network f : Real -> Real
+network f : Tensor Real [1] -> Tensor Real [1]
```
MatthewDaggitt: Same for this unfortunately, its part of the test suite so changing it will cause tests to fail. Please could remove the changes from the PR?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+def generate_loss_function(pathToSpec:str, functionName:str, resources:Dict[str,any]):
+    '''
+    pathToSpec: path to the vehicle spec .vcl file
+    functionName: name of the function for which we want to create the loss function
+    resources: dictionary mapping from the name of the resources declared in the spec to the python implementation
+    '''
+
+    jsonDICT = generateJson(pathToSpec, functionName)
+    context = []
+    loss = '_'
```
MatthewDaggitt: What's the purpose of this line? I think it can be rmeoved?

src/python/vehicle.py
```diff
@@ -0,0 +1,183 @@
...
+
+def generate_loss_function(pathToSpec:str, functionName:str, resources:Dict[str,any]):
+    '''
+    pathToSpec: path to the vehicle spec .vcl file
+    functionName: name of the function for which we want to create the loss function
+    resources: dictionary mapping from the name of the resources declared in the spec to the python implementation
+    '''
+
+    jsonDICT = generateJson(pathToSpec, functionName)
+    context = []
```
MatthewDaggitt: Slightly pedantic, but I think the intent becomes slightly more clear if we called this `empty_context`? We can also declare it after we do the translation to json?

MatthewDaggitt (changes requested):
Added review comments.

One final thing is that I think it would be a good to move the main translation code into a class, and have a method for each term in the language, e.g.:
```python
class LossFunctionTranslation:

  def to_loss_function(resources, jsonDict):
    tag = jsonDICT['tag']
    contents = jsonDICT['contents']

    if   tag == 'Con': translate_constant(contents)
    elif tag == 'Min': translate_min(contents[0], contents[1])
    elif ......

  def translate_constant(value):
    ...

  def translate_min(arg1, arg2):
    ...
```

That would then allow us to subclass the class later and override appropriate methods if we want to change the translation, e.g. when we add support for Pytorch?

MatthewDaggitt:
Also @Tgl70 I should say, that please do open issues for the list of tasks that you think need doing on the `Issues` tab in Github. It's a good way of keeping track of them!

MatthewDaggitt:
As in new tasks not done in this PR, not my review suggestions!

src/python/vehicle.py
Tgl70: I could merge the two like this:
`def result_func(context):`
`      domain = domainloss(context)`
      `minLoss = np.Inf`
      `maxLoss = np.NINF`
      `for _ in range(10):`
          `context.insert(0, random.uniform(domain[0], domain[1]))`
`          if contents[0] == 'All':`
`                        maxLoss = max(maxLoss, bodyloss(context))`
`                    elif contents[0] == 'Any':`
`                        minLoss = min(minLoss, bodyloss(context))`
`          context.pop(0)`
`      return minLoss`
Do you prefer it?

src/python/vehicle.py
Tgl70: I have snake_cased the remaining method, I didn't touch the variable names. Is it fine or do you want them too?
I have type annotated the arguments, what do you mean by the return types?
Something like `return jsonDICT:dict`?

MatthewDaggitt: > do you want them too?

Yes please! See the PEP 8 style guide (https://peps.python.org/pep-0008/#function-and-variable-names)

> Something like return jsonDICT:dict?

You put the return types after the colon in the signature, e.g. `def generateJson(pathToSpec, functionName): Dict[String, String]`

src/python/vehicle.py
MatthewDaggitt: Yup, that seems better!

MatthewDaggitt: Also on Github you can use ``` <newline> code <newline> ``` to type set blocks of code. And also best not to resolve comments before the problem is solved :smile:

src/python/vehicle.py
Tgl70: > You put the return types after the colon in the signature, e.g. `def generateJson(pathToSpec, functionName): Dict[String, String]`

Looking online it should be more like `def generateJson(pathToSpec, functionName) -> Dict[String, String]:`
Also, for `generate_loss_function` and `to_loss_function` I have found the type `Callable` because there does not exist a `function` type.

MatthewDaggitt: Yup, sounds good.

PR #157, opened 2022-09-29
done tests for all the methods

Tgl70:


src/python/test_LossFunctionTranslation.py
```diff
@@ -0,0 +1,127 @@
...
+        self.assertEqual(loss_all, 50)
+        self.assertEqual(loss_any, 12)
+
+    
+    def test_lambda(self):
+        resources = {}
+        json_dict = self.load_json('test_lambda')
+        context = []
+        loss = LossFunctionTranslation().to_loss_function(resources, json_dict)(context)(3)
+        self.assertEqual(loss, 3)
```
MatthewDaggitt, line 115: Should add a few more values here to test that it really is encoding a function rather than simply the constant `3`.

src/python/vehicle.py
```diff
@@ -149,7 +149,8 @@ def result_func(context):
                 elif contents[0] == 'Any':
                     min_loss = min(min_loss, body_loss(context))
                 context.pop(0)
-            return min_loss
+            if contents[0] == 'All': return max_loss 
+            elif contents[0] == 'Any': return min_loss
```
MatthewDaggitt, line 154: Otherwise throw a meaningful error message?

src/python/test_LossFunctionTranslation.py
```diff
@@ -0,0 +1,127 @@
+import unittest
+import json
+from vehicle import LossFunctionTranslation
+from tensorflow import keras
+
+class TestLossFunctionTranslation(unittest.TestCase):
+    def load_json(self, file_name):
+        path_to_json = f'/Users/marcocasadio/Projects/vehicle/src/python/test_json/{file_name}.json'
```
MatthewDaggitt: Hmm so you've got an absolute path here so this won't work on anyone else's computer. You need to make it a relative path!

MatthewDaggitt (commented):
Otherwise looks good!

PR #158, opened 2022-09-30
Improve CI to build Agda and Marabou separately

MatthewDaggitt:
First attempt. Will require multiple pushes


MatthewDaggitt:
Enough progress for now. Merging in.

PR #199, opened 2022-11-01
Restructure project

wenkokke:
Changes:

- Factors out `vehicle`, `vehicle-syntax`, `vehicle-python`, and `vehicle-agda` into their own folders.
- Provides a `Setup.hs` for `vehicle-syntax` which automatically calls BNFC, & removes the need for `build.hs`.
- Refactors the test framework to specify executable tests as data, in `test.json` files.
- Provides a utility for creating executable tests.
- Provides scripts for creating the lists of extra source and doc files.
- Provides a linting check for POSIX shell scripts on commit.

Current shortcomings:

- Removes *all* of `build.hs`. It probably is unnecessary to keep an entire Haskell executable around just to tell you what you still need to install, especially since all build tools aside from Haskell are installed automatically. However, we should probably provide an alternative.
- Removes *most* integration tests. These should still be moved to a top-level project. There is currently one Verify test, but it is disabled since it only works if Marabou is installed.
- Does not update `HACKING.md` or any other documentation.

However, given the fact that the more time passes the more difficult this merge will be, I suggest we merge this ASAP, and create issues for the TODOs, assigned to me. If there's anything you'd consider blocking, @MatthewDaggitt, I'd like to request your help to fix it. 

NOTE: Second PR attempting to resolve the merge conflicts.


MatthewDaggitt:
Okay, here's goes nothing...

PR #207, opened 2022-11-02
Expand CI

wenkokke:
- Build Sphinx docs on commit
- Run stylish-haskell and shellcheck via cabal
- Add editor integration test


wenkokke:
This PR fixes a number of issues, but the toplevel scripts were written under the mistaken assumption that build tool depends are runtime depends... But they are obviously not. The gnarly posix process error is what happens when you pass `callProcess` a bad path.

The solution, I think, is to copy the find-executable-or-give-helpful-advice logic from `build.hs` into these scripts, and remove their build-tool-depends.

wenkokke:
Plus, having learned about Cabal scripts, I guess I should remove the last two POSIX shell scripts and, consequently, the shellcheck script.

PR #214, opened 2022-11-03
Python commandline

Tgl70:
I have implemented Natalia's fix and now my test .vcl files generate .json files. 
Currently, 8 out of 14 tests work. The other do not work yet because of `Lam` and `LBool` errors. Natalia told me that she still need to handle those, so as soon as she implements that I will be able to run those tests.


src/python/command_line.py
```diff
@@ -0,0 +1,69 @@
...
+    for name, location in networks.items():
+        network_list.append('--network')
+        network_list.append(name + ':' + location)
+    return network_list
+
+
+def make_dataset_argument(datasets:dict[str, str]) -> List[str]:
+    dataset_list = []
+    for name, location in datasets.items():
+        dataset_list.append('--network')
```
MatthewDaggitt, line 47: This and `make_parameter_argument` look wrong. It should be `--dataset` not `--network` right? 

More generally, all three methods are very similar. Maybe it should just be one generic function `make_resource_arguments(resources:dict[str, str], arg_name:str)` where `arg_name` is the name of the command line argument?

src/python/test_LossFunctionTranslation.py
```diff
@@ -1,128 +1,156 @@
...
+        functionName = 'tensor'
         resources = {}
-        json_dict = self.load_json('test_tensor')
-        context = [5, 2]
-        loss = LossFunctionTranslation().to_loss_function(resources, json_dict)(context)
-        self.assertEqual(loss.shape, (1,))
-        self.assertEqual(loss, 5)
+        loss = generate_loss_function(path_to_vcl, functionName, resources)
+        self.assertEqual((np.array(loss().shape) & np.array(tf.constant([2, 4, 1, 0]).shape)).all(), True)
+        self.assertEqual((np.array(loss()) & np.array([5, 2, 16, 7])).all(), True)
```
MatthewDaggitt, line 42: What does the `&` operator do here? Why do we need to cast `loss()` to `np.array`?

src/python/vehicle.py
```diff
@@ -1,51 +1,79 @@
...
 
-def generate_loss_function(path_to_spec:str, function_name:str, resources:Dict[str,any]) -> Callable:
+def generate_loss_function(specification:str,
+                            function_name:str,
+                            resources:Dict[str,any],
+                            quantifier_sampling:Dict[str,Callable]=None) -> Callable:
     '''
     pathToSpec: path to the vehicle spec .vcl file
     functionName: name of the function for which we want to create the loss function
     resources: dictionary mapping from the name of the resources declared in the spec to the python implementation
```
MatthewDaggitt, line 15: Missing documentation for new `quantifier_sampling` parameter?

MatthewDaggitt (commented):
Otherwise looks good, and will merge once comments addressed :+1:

MatthewDaggitt:
There's also a merge conflict that you'll need to resolve... Let me know if you need any help.

src/python/test_LossFunctionTranslation.py
Tgl70: The end operator and the np cast are needed because tensors cannot be compared. This is the only way I have found to compare the values and the shape.

src/python/vehicle.py
Tgl70: Updated

src/python/command_line.py
Tgl70: Yep, combined them into one.

src/python/test_LossFunctionTranslation.py
wenkokke: Could you possibly extract that into a function with a name that conveys what you're doing here?

MatthewDaggitt: Yup, something like `def assertTensorsEqual(t1, t2):`

MatthewDaggitt:
Closing this PR. Please reply on the open PR.

PR #234, opened 2022-11-10
Python commandline2

Tgl70:


MatthewDaggitt:
I've disabled or accepted the failing tests, as this is still a better situation than before. So once that last comment is addressed, it should be good to merge in. I'll then open issues for them for Natalia.

Issue #182, opened 2022-10-20
Vehicle compiling loss functions which aren't properties


Tgl70:
Vehicle does not generate the json file sometimes.
It works when there is a `@property` and it does not work when there is not.
Here is the command that failed:
```
vehicle compile --target LossFunction --specification ./src/python/test_json/test_maximum.vcl --outputFile ./src/python/t.json
```

MatthewDaggitt:
Hmm can't seem to reproduce this locally. Can you provide the contents of `test_maximum.vcl` file here?

Tgl70:
The file is now pushed on the branch python-commandline and here is the content:
```
maximum : Bool
maximum = True or False
```

MatthewDaggitt:
Hmm still struggling to reproduce this locally, even on the `python-commandline` branch running the exact same python code... I have pushed a version of your branch up to the repo which has been merged into the latest version of the `dev` branch, so you definitely want to pull the latest of `python-commandline`. Can you check that you still get the same problem now?

One thing I did spot is that in `call_vehicle_to_generate_loss_json` it should be:
```
path_to_json_directory + '/loss_function.json'
``` 
rather than 
```
path_to_json_directory + 'loss_function.json'
```
Not sure why that would make a difference but...

MatthewDaggitt:
@ndslusarz and I managed to diagnose this as the incorrect `isProperty` check in the compilation of loss functions.

MatthewDaggitt:
This was fixed in https://github.com/vehicle-lang/vehicle/pull/254

PR #247, opened 2022-11-15
Multiple differentiable logics

ndslusarz:
This adds support for multiple different translations of constaints for loss functions (differentiable logics) that the user can choose. The tests have been updated to check all of them.

It also adds handling Lambdas, Lets and booleans now to the translation.

Some elements got added to LExpr (Lam, Let and Power) will need to be handled by python side later. Fixed the issue #182 .


MatthewDaggitt:
Thanks! Unfortunately it looks like your rebase after @wenkokke's project changes hasn't quite worked as I assume you didn't mean to change 317 files!

Are you happy sorting it out or would you like some help?

ndslusarz:
Just checked, definitely some files that I haven't touched I must have pushed by mistake - the ones in gen.hs.Vehicle generally -  (shouldn't be affected by the merge though  as I started this branch fresh after that, but maybe something went wrong when pulling) I'll go and fix those today.

A lot of the files though are from tests since I was running them for all DLs (there is 5 of them and each gets their own json, err, ...). I'm not sure if you want to keep all the tests in or would you prefer to have them run just for one translation?

MatthewDaggitt:
Oh wow you're right. Yes please fix the ones that aren't supposed to be touched. 

That is a lot of tests! I think as the translations only differ by how they treat the logical parts, I would come up with a single artificial test spec that uses all the fields of `Translation` and run all the different translations on that one. That should get us 100% test coverage on the individual translations? Then for the rest of the tests, I would only run them for the default translation (whatever you choose that to be).

Does that sound like a plan? (Let me know if I've missed something!)

ndslusarz:
Good idea, I'll do that for the tests, this way I'm not adding that many but it still gets tested once in case something broke. I'll add some sort of artificial test like you said that goes through every field that's dependant on DL and can hypothetically break (all the other fields don't change so they don't necessarily need to be tested on all translations, just one, only comparisons, logical operators and booleans are dependant on translations). Default is DL2 so I'll leave the tests for that one. 

Other files should be removed by now, it seems this folder no longer exists in dev so they were probably moved in restructuring and I didn't pull something on my end.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +52,74 @@ newtype Domain = Domain ()
 instance FromJSON Domain
 instance ToJSON Domain
 
---definitoon of the LExpr - all expressions allowed in loss constraint
+--definitoon of the LExpr - all expressions allowed in a loss constraint
```
MatthewDaggitt: Typo `definitoon`

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +52,74 @@ newtype Domain = Domain ()
...
+--they can be found in Vehicle.Backend.Prelude and the default option if none is provided is DL2.
 
-compileProg :: MonadCompileLoss m => V.CheckedProg -> m [LDecl]
-compileProg  (V.Main ds) = catMaybes <$> traverse compileDecl ds
+compile :: MonadCompile m => DifferentiableLogic -> V.CheckedProg -> V.PropertyContext -> NetworkContext -> m [LDecl]
+compile d prog propertyCtx networkCtx = do
+  normalisedProg <- normalise prog normalisationOptions
+  case d of
+    DL2 -> do
+      runReaderT (compileProg dl2Translation normalisedProg) (propertyCtx, networkCtx)
```
MatthewDaggitt: We could get rid of a lot of the duplication if we had a standalone function of the type `DifferentiableLogic -> Translation` and then used it here?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +52,74 @@ newtype Domain = Domain ()
...
+      runReaderT (compileProg godelTranslation normalisedProg) (propertyCtx, networkCtx)
+    Lukasiewicz -> do
+      runReaderT (compileProg lukasiewiczTranslation normalisedProg) (propertyCtx, networkCtx)
+    Product -> do
+      runReaderT (compileProg productTranslation normalisedProg) (propertyCtx, networkCtx)
+    Yager -> do
+      runReaderT (compileProg yagerTranslation normalisedProg) (propertyCtx, networkCtx)
+
+
+--compile entire specification (calls compileDecl)
```
MatthewDaggitt: Function comments should start with a `--| ` so they are picked up by the Haskell documentation generator.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +128,83 @@ compileDecl d =
...
+
+--compile argument
+compileArg :: MonadCompile m => Translation -> V.CheckedArg -> m LExpr
+compileArg t arg = compileExpr t (V.argExpr arg)
 
-compileLiteral :: V.Literal -> Double
-compileLiteral = \case
-  V.LUnit{}    -> developerError "LUnit"
-  V.LBool{}    -> developerError "LBool"
+--helper function for compiling Lieterals
```
MatthewDaggitt: Typo `Lieterals`

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +128,83 @@ compileDecl d =
...
 currentPass :: Doc a
 currentPass = "compilation to loss functions"
 
-compileArg :: MonadCompile m => V.CheckedArg -> m LExpr
-compileArg arg = compileExpr (V.argExpr arg)
+--part of the syntax translation that differ depending on chosen DL namely:
+    -- logical connectives (not, and, or, implies)
+    -- comparisons (<, <=, >, >=, =, !=)
+
+--compile argument
```
MatthewDaggitt: Not sure this documentation comment adds any new information? Maybe remove?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +128,83 @@ compileDecl d =
...
+    return (Just (DefFunction (nameOf ident) expr'))
 
 currentPass :: Doc a
 currentPass = "compilation to loss functions"
 
-compileArg :: MonadCompile m => V.CheckedArg -> m LExpr
-compileArg arg = compileExpr (V.argExpr arg)
+--part of the syntax translation that differ depending on chosen DL namely:
+    -- logical connectives (not, and, or, implies)
+    -- comparisons (<, <=, >, >=, =, !=)
```
MatthewDaggitt: This documentation should go on the `Translation` object?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
 compileQuant V.Forall = All
 compileQuant V.Exists = Any
 
+data Translation = Translation
```
MatthewDaggitt: Would `DifferentialLogicImplementation` be a better name for this record?

Also maybe we should move this record, and all the translation objects to a separate file? `LossFunction.Logics`?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+    { compileAnd          :: LExpr -> LExpr -> LExpr,
+      compileOr           :: LExpr -> LExpr -> LExpr,
```
MatthewDaggitt: We usually layout records, by putting the comma on the start of the new line aligned with the `{` brace. See everywhere else in the compiler for examples.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+      compileImplication  :: LExpr -> LExpr -> LExpr,
+
+      compileLe           :: LExpr -> LExpr -> LExpr,
+      compileLt           :: LExpr -> LExpr -> LExpr,
+      compileGe           :: LExpr -> LExpr -> LExpr,
+      compileGt           :: LExpr -> LExpr -> LExpr,
+      compileEq           :: LExpr -> LExpr -> LExpr,
+      compileNeq          :: LExpr -> LExpr -> LExpr,
+
+      compileBool         :: Bool -> Double
```
MatthewDaggitt: We could avoid a `case` in every instantiation by having a `compileTrue` and a `compileFalse`?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+     compileBool = \case
+                      True -> 1
+                      False -> 0
+   }
+
+yagerTranslation :: Translation
+yagerTranslation = yagerTranslationCompile 1 --change constant here
+
+--helper function that sets parameter p for the Yager DL (by default set to 1)
+yagerTranslationCompile :: Rational -> Translation
```
MatthewDaggitt: Maybe `parameterisedYagerTranslation`?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+
+     compileBool = \case
+                      True -> 1
+                      False -> 0
+   }
+
+yagerTranslation :: Translation
+yagerTranslation = yagerTranslationCompile 1 --change constant here
+
+--helper function that sets parameter p for the Yager DL (by default set to 1)
```
MatthewDaggitt: We should have the documentation "by default set to 1" on the `yagerTranslation`, not this definition right?

vehicle/src/Vehicle/CommandLine.hs
```diff
@@ -192,8 +192,11 @@ allITPs = map show (enumerate @ITP)
 allVerifiers :: [String]
 allVerifiers = map show (enumerate @VerifierIdentifier)
 
+allLossFunctionDLs :: [String]
+allLossFunctionDLs = map show (enumerate @DifferentiableLogic)
+
 allBackends :: [String]
-allBackends = allVerifiers <> allITPs <> map show [LossFunction, TypeCheck]
+allBackends = allVerifiers <> allITPs <> allLossFunctionDLs --map show [LossFunction, TypeCheck]
```
MatthewDaggitt: Could remove the comment here? (I know you didn't add it!)

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+
+  V.FromNat {}       -> True
+  V.FromRat {}       -> True
+  V.FromVec {}       -> True
+  V.Foreach{}        -> True
+
+  _                  -> False
+
+-----------------------------------------------------------------------
+-- debugging options
```
MatthewDaggitt: Missing an empty line here? (very minor I know!)

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
 compileQuant V.Forall = All
 compileQuant V.Exists = Any
 
+data Translation = Translation
+    { compileAnd          :: LExpr -> LExpr -> LExpr,
```
MatthewDaggitt: Identation of `{` should be 2 spaces.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +213,197 @@ compileQuant :: V.Quantifier -> Quantifier
...
+
+---------------------------------------------------------------------------------------
+-- different differentiable logic translations (options were listed before. DL2 is the default translation)
+
+dl2Translation :: Translation
+dl2Translation = Translation
+   { --double check implication, do negation properly
+     compileAnd = Addition,
+     compileOr  = Multiplication,
+     compileNot = id, --this should be normalised out and pushed to the innermost level of comparisons by now
```
MatthewDaggitt: This is the only thing that worries me in a major way. If that particular normalisation is not done here, then that means that the negation is normalised out for every logic, not just DL2 right? We don't want that?

However, that may mean that the translation will have to accept terms of type `CheckedExpr` rather than `LExpr` which sucks...

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +128,83 @@ compileDecl d =
...
   V.LNat     e -> fromIntegral e
   V.LInt     e -> fromIntegral e
   V.LRat     e -> fromRational e
 
-compileExpr :: MonadCompile m => V.CheckedExpr -> m LExpr
-compileExpr e = showExit $ do
+--helps compile a name from DBBinding, even if there is no name given
+compileDBBinding :: Maybe Name -> Name
+compileDBBinding =
+  fromMaybe "No_name"
```
MatthewDaggitt: This could go on the previous line?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +52,74 @@ newtype Domain = Domain ()
...
+--------------------------------------------------------------------------------
+-- different avilable differentiable logics (types of translation from the constraint to loss function) are
+-- DL2
+-- Godel
+-- Lukasiewicz
+-- Product based
+-- Yager
 
-  _                  -> False
+--they can be found in Vehicle.Backend.Prelude and the default option if none is provided is DL2.
```
MatthewDaggitt: Again, these comments should go next to the translations rather than here right?

MatthewDaggitt (commented):
Looks great in general! I have a variety of minor comments and one major one. Also there's two rogue files being added in the `.shake/` directory!

MatthewDaggitt:
Ah also in the latest version of `dev`, `Vehicle.Language.AST.Core` has been renamed `Vehicle.Language.AST.Expr` which I think is why the CI tests are failing?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: That's one I worried about as well, but came to the same conclusion and didn't want to start passing `CheckedExpr ` into translations... I'll fix the other issues for now and think about this. In theory normalising the nagtion shouldn't harm the other translations but then they don't need it so it's adding unnecessary complications... 

The only other option I can see is doing some sort of case split on `NotExpr` to have it have different behaviour depending on translation (`compileNot` for all the other ones and the additional negation normalisation for DL2 only?)

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Tried moving it, but then I have to import things such as LExpr from LossFunction.Compile and it errors out because of cyclic module dependency between these two. Is there any solution to that?

ndslusarz:
Thanks for the feedback! Most of the comments have been resolved in the latest push (aside from the few I commented on, I marked the others as resolved). 

I've changed the name to Vehicle.Language.AST.Expr which should hopefully fix the tests (I didn't know there was another restructuring/renaming of files after the last one which I pulled) - though this caused some conflicts since I had to correct the name everywhere and I assume it looks slightly different to dev (I'll take a look at them but probably tomorrow).

New test for all DLs not there yet.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: in one of the latest commits

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: in one of the latest commits

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Does the `case` impact it badly? It seems more readable to me with the `compileBool` but I can swap it to `compileTrue` and `compileFalse` if you prefer that.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -1,22 +1,27 @@
 module Vehicle.Backend.LossFunction.Compile
   ( LDecl
+  ,DifferentiableLogic
```
MatthewDaggitt: Space between comma and the name?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +53,63 @@ newtype Domain = Domain ()
 instance FromJSON Domain
 instance ToJSON Domain
 
---definitoon of the LExpr - all expressions allowed in loss constraint
+-- definitoon of the LExpr - all expressions allowed in a loss constraint
```
MatthewDaggitt: The typo is still here :wink:

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +53,63 @@ newtype Domain = Domain ()
...
 
-    _              -> False
-
-  V.FromNat {}       -> True
-  V.FromRat {}       -> True
-  V.FromVec {}       -> True
-  V.Foreach{}        -> True
+chooseTranslation :: DifferentiableLogic -> DifferentialLogicImplementation
+chooseTranslation =
+  \case
```
MatthewDaggitt: We could save a line and an indentation level by putting the case on the previous line?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +197,203 @@ compileQuant :: V.Quantifier -> Quantifier
...
+  V.FromVec {}       -> True
+  V.Foreach{}        -> True
+
+  _                  -> False
+
+data DifferentialLogicImplementation = DifferentialLogicImplementation
+    { compileAnd          :: LExpr -> LExpr -> LExpr
+    , compileOr           :: LExpr -> LExpr -> LExpr
+    , compileNot          :: LExpr -> LExpr
+    , compileImplication  :: LExpr -> LExpr -> LExpr
```
MatthewDaggitt: I think `Implication` is called `Implies` elsewhere in the compiler. It's also a bit shorter so this field will stick out less. The indentation should also be 2 spaces here rather than 4 spaces.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -48,87 +53,63 @@ newtype Domain = Domain ()
...
 
 type MonadCompileLoss m =
   ( MonadCompile m
   , MonadReader (V.PropertyContext, NetworkContext) m
   )
 
-compileDecl :: MonadCompileLoss m => V.CheckedDecl -> m (Maybe LDecl)
-compileDecl d =
+-- |compile all functions found in spec, save their names (call compileExpr on each)
+compileDecl :: MonadCompileLoss m => DifferentialLogicImplementation -> V.CheckedDecl -> m (Maybe LDecl)
```
MatthewDaggitt: Now that we compile everything, I don't think this function needs to return a `Maybe LDecl` anymore? It can just return `LDecl`?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +118,77 @@ compileDecl d =
...
+compileExpr :: MonadCompile m => DifferentialLogicImplementation -> V.CheckedExpr -> m LExpr
+compileExpr t e = showExit $ do
   e' <- showEntry e
   case e' of
     --logical operatives
-    V.NotExpr     _ [e1]     -> Negation <$> compileArg e1
-    V.AndExpr     _ [e1, e2] -> Min <$> compileArg e1 <*> compileArg e2
-    V.OrExpr      _ [e1, e2] -> Max <$> compileArg e1 <*> compileArg e2
-    V.ImpliesExpr _ [e1, e2] -> Max <$> (Negation <$> compileArg e1) <*> compileArg e2
+    V.NotExpr     _ [e1]     -> compileNot t <$> compileExpr t (lowerNot (argExpr e1))
```
MatthewDaggitt: Hmm so looking at this again, this is actually a bug in the other translations. `lowerNot` actually eliminates the `not` by pushing it in, so it is incorrect to call `compileNot` on the result, as there is no `not` in the program anymore!

See original comment chain for discussion on what we should do.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +118,77 @@ compileDecl d =
...
+        V.Lt -> compileLt t <$> compileArg t e1 <*> compileArg t e2
+        V.Ge -> compileGe t <$> compileArg t e1 <*> compileArg t e2
+        V.Gt -> compileGt t <$> compileArg t e1 <*> compileArg t e2
+
+    V.VecLiteral _ _ xs                -> TensorLiteral <$> traverse (compileExpr t) xs
+    V.Literal _ l                      -> return $ Constant $ compileLiteral t l
+    V.App _ (V.Var _ (V.Free ident)) p -> NetworkApplication (V.nameOf ident) <$> traverse (compileArg t) p
+    V.Var _ (V.Bound var)              -> return (Variable var)
+    V.AtExpr _ _ _ [xs, i]             -> At <$> compileArg t xs <*> compileArg t i
+    V.Let _ x name expression          -> Let (compileDBBinding (V.binderRepresentation name)) <$> compileExpr t x <*> compileExpr t expression
```
MatthewDaggitt: Could we rename `name` in this and the case below to `binder`? They are actually binder objects, not just names. (Binders contain a whole lot of additional information apart from just the name!)

vehicle/tests/golden/compile/mnist-robustness/TypeCheck.err.golden
```diff
@@ -0,0 +1 @@
+syntax error at line 1, column 1 before `.'
```
MatthewDaggitt: Err I'm a bit worried about this. Why is the `mnist-robustness` test now producing an error message when we're type checking? Is this genuine? (i.e. if you delete this file and rerun the test does it generate the same error).

I think there's the same issue in `windController` test.

MatthewDaggitt (commented):
A few more comments.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
MatthewDaggitt: I think its clearer, as `compileBool` might be confused as compiling the type `Bool` (which we do in other backends (though I'll probably now find I've got other `compileBool` functions elsewhere :sweat_smile: )).  I think `compileTrue` and `compileFalse` are more accurate (and more readable!).

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
MatthewDaggitt: Hmm okay, so I think the answer is to have the type of the `compileNot` field in the `Translation` record to have the type `Maybe (LExpr -> LExpr -> LExpr)` rather than `LExpr -> LExpr -> LExpr`. A `Nothing` value then indicates that it doesn't exist, and that the `not` should be pushed through. Then to use it, you have to case split on `compileNot` first.

Thoughts?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
MatthewDaggitt: I think the answer is that we should move `LExpr` in there as well?

MatthewDaggitt:
I've also tried to resolve some of the merge problems so you'll need to pull this morning :+1:

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +197,203 @@ compileQuant :: V.Quantifier -> Quantifier
...
+-- Product based
+-- Yager
+
+--they can be found in Vehicle.Backend.Prelude and the default option if none is provided is DL2.
+
+-- part of the syntax translation that differ depending on chosen DL are:
+    -- logical connectives (not, and, or, implies)
+    -- comparisons (<, <=, >, >=, =, !=)
+
+dl2Translation :: DifferentialLogicImplementation
```
MatthewDaggitt: Oh and it might be nice to document which paper you took each translation from in a comment on the translation?

MatthewDaggitt:
See the failing tests as well.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: But moving the `LExpr` will require me to explicitly export all elements of it (since they are used in `compileExpr`, the main function there) and makes it less readable since the traslation of the `LExpr` is in the `Compile` file, won't it? Plus if I move that I have to move with it definitions of `Quantifier,` `Domain` - if you want to split it still I'm happy to do that, but might make it harder to find and/or edit what is being translated if someone else takes a look at it.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Sure, will do.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
MatthewDaggitt: > But moving the LExpr will require me to explicitly export all elements of it (since they are used in compileExpr, the main function there) 

You can export all constructors in a single line by going `LExpr (..)`

> Plus if I move that I have to move with it definitions of Quantifier, Domain

Yes you would.

> makes it less readable since the traslation of the LExpr is in the Compile file, won't it?

In a large project, its kind of inevitable that things are going to be split across multiple files... `CheckedExpr` isn't in the same file right?

Long story short, its up to you. You can move it if you want to or leave it where it is :+1:

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Alright, this sounds sensible, will try that.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Done in one of the latest commits 👍

vehicle/tests/golden/compile/mnist-robustness/TypeCheck.err.golden
ndslusarz: I've done that and the problem stays. It's actually a problem for mnist-robustness, windController, acasXu tests (it seems I haven't pushed the .err files for this one). The tests look the same as in the dev branch, no changes there. For me the .vcl files for these contain a path to the actual spec in /examples (let me know if that shouldn't be the case, it seems some of my problems came from my old pull not deleting certain folders that it should have) - my theory for now is that it can't read them, it errors out on the dot. I've got no idea yet about why they fail for me if they work in general... I'll take a closer look...

MatthewDaggitt: Ah it's a symlink problem. Summoning @wenkokke! Wen, you said that there shouldn't be any problems with symlinks on Windows, but it looks like there is?

For the meantime @ndslusarz, can you avoid committing these error files and rely on the CI on Github for testing them?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +159,54 @@ compileQuant :: V.Quantifier -> Quantifier
 compileQuant V.Forall = All
 compileQuant V.Exists = Any
 
+
```
MatthewDaggitt: Can we have a single empty line between declarations (and comments) please?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -217,6 +159,54 @@ compileQuant :: V.Quantifier -> Quantifier
...
+----------------------------------------------------------------------------------------------------
+--handling normalisation options
```
MatthewDaggitt: A space between the `--` and the first letter?

Also can we capitalise the first letters of comments and headers?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+
+
+instance FromJSON Domain
+instance ToJSON Domain
+
+instance FromJSON LExpr
+instance ToJSON LExpr
+
+instance FromJSON Quantifier
+instance ToJSON Quantifier
```
MatthewDaggitt: These instances should live underneath the declaration of each datatype rather than all at the top.

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+
+-- definiton of the LExpr - all expressions allowed in a loss constraint
```
MatthewDaggitt: Again will need `|` to make the documentation pick this comment up?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+  | Quantifier Quantifier Name Domain LExpr   -- quantifiers forall, exists
+  | At LExpr LExpr                            -- at
+  | TensorLiteral [LExpr]                     -- tensor
+  | Lambda Name LExpr                         -- lambda expression
+  | Let Name LExpr LExpr                      -- let expression
+  | Power LExpr LExpr                         -- exponential
+  deriving (Eq, Ord, Generic, Show)
+
+--------------------------------------------------------------------------------
+-- other Definitions
```
MatthewDaggitt: Weird capitalisation?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+  , compileEq           :: LExpr -> LExpr -> LExpr
+  , compileNeq          :: LExpr -> LExpr -> LExpr
+
+  , compileTrue         :: Double
+  , compileFalse        :: Double
+    }
+
+
+--------------------------------------------------------------------------------
+-- different avilable differentiable logics (types of translation from the constraint to loss function) are
```
MatthewDaggitt: Could we make sure we don't massively overrun the line lengths here with the comments please? (The header lines are indicators of where they should stop).

Also typo in `avilable`

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+-- DL2                              --Fischer, Marc, et al. "Dl2: Training and querying neural networks with logic."  PMLR, 2019.
```
MatthewDaggitt: Could we add these as documentation comments on each translation rather than as free floating comments?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+-- Godel                            --van Krieken, et al. "Analyzing differentiable fuzzy logic operators." Artificial Intelligence 302 (2022): 103602.
+-- Lukasiewicz                      --as above
+-- Product based                    --as above
+-- Yager                            --as above
+
+--they can be found in Vehicle.Backend.Prelude and the default option if none is provided is DL2.
```
MatthewDaggitt: I'm not sure documenting this here is that useful. Is this documented on the command line interface? That would be a better place for it?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+
+data Quantifier
+  = All
+  | Any
+  deriving (Eq, Ord, Generic, Show)
+
+newtype Domain = Domain ()
+  deriving (Eq, Ord, Generic, Show)
+
+-- template for different avilable differentiable logics
```
MatthewDaggitt: Again should be a doc comment?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+-- Godel                            --van Krieken, et al. "Analyzing differentiable fuzzy logic operators." Artificial Intelligence 302 (2022): 103602.
+-- Lukasiewicz                      --as above
+-- Product based                    --as above
+-- Yager                            --as above
+
+--they can be found in Vehicle.Backend.Prelude and the default option if none is provided is DL2.
+
+-- part of the syntax translation that differ depending on chosen DL are:
+    -- logical connectives (not, and, or, implies)
+    -- comparisons (<, <=, >, >=, =, !=)
```
MatthewDaggitt: Add this to the doc comment for `DifferentialLogicImplementation`?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+  , compileLe           :: LExpr -> LExpr -> LExpr
+  , compileLt           :: LExpr -> LExpr -> LExpr
+  , compileGe           :: LExpr -> LExpr -> LExpr
+  , compileGt           :: LExpr -> LExpr -> LExpr
+  , compileEq           :: LExpr -> LExpr -> LExpr
+  , compileNeq          :: LExpr -> LExpr -> LExpr
+
+  , compileTrue         :: Double
+  , compileFalse        :: Double
+    }
```
MatthewDaggitt: Misaligned `}`

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+  = All
+  | Any
+  deriving (Eq, Ord, Generic, Show)
+
+newtype Domain = Domain ()
+  deriving (Eq, Ord, Generic, Show)
+
+-- template for different avilable differentiable logics
+
+data DifferentialLogicImplementation = DifferentialLogicImplementation
```
MatthewDaggitt, line 71: This might sit better under the differential logics header below?

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+
+-- part of the syntax translation that differ depending on chosen DL are:
+    -- logical connectives (not, and, or, implies)
+    -- comparisons (<, <=, >, >=, =, !=)
+
+dl2Translation :: DifferentialLogicImplementation
+dl2Translation = DifferentialLogicImplementation
+  { compileAnd = Addition
+  , compileOr  = Multiplication
+  , compileNot = Nothing --this should be normalised out and pushed to the innermost level of comparisons by now
```
MatthewDaggitt: This comment should be attached to the `compileNot` field in the `DifferentialLogicImplementation` declaration

vehicle/src/Vehicle/Backend/LossFunction/Logics.hs
```diff
@@ -0,0 +1,210 @@
...
+instance FromJSON LExpr
+instance ToJSON LExpr
+
+instance FromJSON Quantifier
+instance ToJSON Quantifier
+
+-- definiton of the LExpr - all expressions allowed in a loss constraint
+
+data LExpr
+  = Negation LExpr                           -- this is minus, not the logical operation of negation
```
MatthewDaggitt: You can attach documentation comments to individual constructors as follows:
```haskell
= Negation LExpr
-- |^ this is minus, not the logical operation of negation
```

Also I'm not sure if some of these comments add very much. For example the `Multiplication -- multiplication` really doesn't help. Most of these apart from the negation comment and perhaps the exponential comment fall into this category I think?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +73,84 @@ compileDecl d =
...
   e' <- showEntry e
   case e' of
     --logical operatives
-    V.NotExpr     _ [e1]     -> Negation <$> compileArg e1
-    V.AndExpr     _ [e1, e2] -> Min <$> compileArg e1 <*> compileArg e2
-    V.OrExpr      _ [e1, e2] -> Max <$> compileArg e1 <*> compileArg e2
-    V.ImpliesExpr _ [e1, e2] -> Max <$> (Negation <$> compileArg e1) <*> compileArg e2
+    V.NotExpr     _ [e1]     -> case compileNot t of
+                                  Nothing -> compileExpr t (lowerNot (argExpr e1))--compileNot t <$> compileExpr t (lowerNot (argExpr e1))
+                                  Just f -> f <$> compileArg t e1
```
MatthewDaggitt: Delete remaining comment, and we should probably decrease the indentation of the cases to only 2 spaces beyond the `V.NotExpr`

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -137,78 +73,84 @@ compileDecl d =
...
+  V.LInt     e     -> fromIntegral e
+  V.LRat     e     -> fromRational e
+
+-- |helps compile a name from DBBinding, even if there is no name given
+compileDBBinding :: Maybe Name -> Name
+compileDBBinding = fromMaybe "No_name"
+
+--compileMaybeNot :: m LExpr -> m LExpr
+--compileMaybeNot not = case not of
+--  Nothing -> (lowerNot (argExpr e1))
```
MatthewDaggitt: Delete?

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -30,105 +35,36 @@ data LDecl
...
-normalisationOptions = Options
-  { declContext                 = mempty
-  , boundContext                = mempty
-  , normaliseDeclApplications   = False
-  , normaliseLambdaApplications = False
-  , normaliseStdLibApplications = False
-  , normaliseBuiltin            = normBuiltin
-  , normaliseWeakly             = False
-  }
+chooseTranslation :: DifferentiableLogic -> DifferentialLogicImplementation
```
MatthewDaggitt: This should probably go in the `Logics` file?

MatthewDaggitt (commented):
Thanks, looks pretty much there functionality wise and major comments wise! I've add a load of tiny, niggling format comments but those will be the last ones, promise!

MatthewDaggitt:
@ndslusarz all looks good apart from two failing Python tests. Can you liaise with @Tgl70 and decide what the correct answer for these tests should be?

After that I'll hit merge :+1: 

(Note: I have merged `dev` into this branch so you'll need to pull locally before making any more edits)

Tgl70:
> 

@MatthewDaggitt  the correct answer for the first test is actually 1 (the same that the test is expecting) that's why ```AssertionError: 0 != 1```.
While the second test seems to fail because it generates an empty json (https://github.com/vehicle-lang/vehicle/blob/dev/vehicle-python/tests/data/test_quantifier_all.json).
Not sure why because both worked for me.

ndslusarz:
The second one is on me, I forgot about the bounded forall, I'll try and add that.

@MatthewDaggitt Not sure about the first one. The .json for that file looks outdated and not from this branch, the translation used there is old which you can see since it doesn't change between dev and this branch and the translation for >= used to be different. (@Tgl70 said the jsons used to not be saved so I'm not sure if they get overwritten or if it's from an old branch)
 Based on this json that I can see in https://github.com/vehicle-lang/vehicle/blob/dev/vehicle-python/tests/data/test_network.json the answer should be the one from the test and I can't see any errors there (but I also can't see the updated json).

MatthewDaggitt:
Okay, I think the priority is getting this merged ASAP. If you can fix the first test quickly @ndslusarz then I guess do so! Otherwise, I recommend we disable one or both of them, merge it, and then open an issue for the remaining bugs?

ndslusarz:
Alright, just figured out the `AssertionError: 0 != 1` - it should actctually expect 0 as this is a DL2 translation so if the constraint was satisfied it should in fact return 0 as that's its True value. I corrected that now and pushed, it should fix the issue I hope (if it doesn't I'll disable this test temporarily as well).
 I disabled the test for quantifiers, I'll add that later separately then in case it takes me longer to implement.

One thing that remains is that I see only old version of jsons in that folder of python tests, they didn't seem to get updated with my new translation but that's tangential to my translation.

MatthewDaggitt:
> One thing that remains is that I see only old version of jsons in that folder of python tests, they didn't seem to get updated with my new translation but that's tangential to my translation.

Yes, I'm not sure why they exist, as at the moment the JSON is regenerated automatically by each test (although see #268 for why that might change soon). I wouldn't worry about it now :+1:

PR #248, opened 2022-11-17
Fix #246 by improving the error message for invalid property types

### LINKED ISSUE #246: More meaningful error for wrong type on @Property
Anything labelled `@property` in a spec needs to be of the type Bool. Currently when you have a function with some other type and label it a `@property` you get this error message:

```
Error at Line 13, Columns 1-13: Invalid type 'Label -> Label -> Bool' for property 'isOneOrOther'.
Fix: change the type of 'isOneOrOther' to one of `Bool`, `Vector Bool n` or `Tensor Bool n`.
```

which is not very helpful when identyfyig the cause. Maybe something like `@property expected to be of type Bool. The actual type is ...` or something similiar to give clear indication of what the cause is?

MatthewDaggitt:
The error message is now:
```
Error at Line 13, Columns 1-13: The type of @property 'isOneOrOther':
  Label -> Label -> Bool
is not supported. Only the following types are allowed for '@property's:
  Bool, Vector Bool n, Tensor Bool ns
Fix: either change the type of 'p' to a supported type or remove the '@property' annotation.
```
@ndslusarz better?

### DIFF
--- a/vehicle/src/Vehicle/Compile/Error.hs
+++ b/vehicle/src/Vehicle/Compile/Error.hs
@@ -120,7 +120,7 @@ data CompileError
   | InferableParameterContradictory      Identifier (DeclProvenance, Resource, Int) (DeclProvenance, Resource, Int)
   | InferableParameterUninferrable       DeclProvenance
 
-  | PropertyTypeUnsupported         DeclProvenance CheckedType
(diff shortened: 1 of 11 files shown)


MatthewDaggitt:
Drat, sorry, merged this without thinking.

PR #273, opened 2022-11-21
Support GHC 9.4

wenkokke:


.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           - [ubuntu-latest, Linux]
           - [macos-latest, macOS]
           - [windows-latest, Windows]
-        ghc-version: ["8.10", "9.0", "9.2"]
+        ghc-version: ["8.10.7", "9.0.2", "9.2.4", "9.4.2"]
```
MatthewDaggitt, line 26: Why this change out of interest?

wenkokke: Freeze files are specific to minor versions, so the CI fails if there's a minor version mismatch. GHC 9.2 and 9.4 currently resolve to 9.2.4 and 9.4.3, instead of the latest versions.

PR #279, opened 2022-11-21
Set fail-fast: true

wenkokke:


MatthewDaggitt:
Related question: are we hammering the Github servers too much with all this CI? What are our limits?

PR #285, opened 2022-11-22
Set mypy type checking to strict

wenkokke:
Set mypy type checking to strict, which does strict type checking.

This PR still needs the actual work to make the Python code typed.

@Tgl70 Could you help with that?

If it helps, I think the type of JSON objects is:
```
Json = Union[Dict[str, Json], List[Json], str, int, float, bool, None]  # type: ignore
```
Note that requires a mypy pragma to pass, since mypy doesn't support recursive types.

However, PyLance does, so your editor support will probably be fine.


wenkokke:
Closed in favour of #344.

PR #286, opened 2022-11-22
Run cabal-fmt & stylish-haskell

wenkokke:


wenkokke:
@MatthewDaggitt I think your pre-commit installation may be broken?

MatthewDaggitt:
So both cabal format and stylish haskell are running on my pre-commit hooks. And cabal format definitely makes changes. Stylish haskell does not however.... Maybe we should diagnose next time we're working together?

PR #291, opened 2022-11-24
now all the tests work after merging with Natalia

Tgl70:
The previous tests all work now.
I have added Power but cannot test it because I cannot generate it from a vcl spec with the DL2 translation.
The same happens tho, for the Minimum function.
Need to add a translation for Let.
Now, I see 3 options:
- we do not test those, as we already test other math functions and Maximum
- we test those by generating them with different translations
- we either swap all tests or just test these functions by manually creating json tests instead of creating them from vcl specs.


MatthewDaggitt:
Thanks, looks great.

> I have added Power but cannot test it because I cannot generate it from a vcl spec with the DL2 translation.

I think Natalia has added an option to the command line so that you can specify which translation you want to use. Can you check with her how to use it? We should then add an (optional) Python enum argument to the main function to allow the user to choose?

As you can see at the bottom of this PR,  Wen has also installed some new automatic "pre-commit" tests that help flag up issues with code. I think it's currently failing because we have no `__init__.py` module in the directory. Any chance you could add one?

(You can install the pre-commit hooks locally so that they automatically run before every commit you make [here](https://github.com/vehicle-lang/vehicle/blob/dev/HACKING.md#pre-commit-hooks))

wenkokke:
@Tgl70 I saw that you added some code to change the Python load path. Please use poetry to run the tests going forwards. I've left more instructions in the README for vehicle-python, which I'll add here as well:

To install the dependencies _and_ the development dependencies, run the following command from vehicle-python:

```sh
poetry install --no-root --with dev
```

To run the loss function tests, run the following command from vehicle-python:

```sh
poetry run tox
```

To run the training test, run the following command from vehicle-python:

```sh
poetry run python tests/test.py
```

wenkokke:
I've also added a meaningful error message for when vehicle isn't available on the path.

PR #308, opened 2022-12-02
Load empty standard library

MatthewDaggitt:
With flag to disable loading....


MatthewDaggitt:
@wenkokke these tests are failing because Vehicle can't find the environment variables inside the TOX virtual environment. Any ideas?

The only one I have is to fall back to adding the standard library file next to the specification file...

MatthewDaggitt:
Okay, hopefully worked out how to get the environment variables passed through tox

PR #315, opened 2022-12-02
Convert AST to be strict and derive instances of NoThunks

wenkokke:
This PR makes _some_ of the AST strict, but there are still several major sources of thunks, specifically:

- the use of `List` in `GenericProg.Main` and `Expr.LVec`;
- the use of `NonEmpty` in `Expr.App`; and
- the use of `Maybe` in `InputBinding`.

The first two should be replaced by `Vector`, either from [`strict-containers`](https://hackage.haskell.org/package/strict-containers) which provides a value-strict vector, or from [`vector`](https://hackage.haskell.org/package/vector) which is perhaps a better maintained package, but only provides structure-strict vectors, which would require us to take care to enforce value-strictness.

For the last we can use `Maybe` from [`strict`](https://hackage.haskell.org/package/strict) as a drop-in replacement.

However, this PR contains a lot of instrumentation for measuring thunks, and therefore should probably be merged.

The NoThunks instrumentation is hidden behind a compiler flag, and therefore behind a CPP macro. We could perhaps do this more cleanly, but I don't necessarily want to have our production binaries depend on debugging only packages.

Unfortunately, to derive NoThunk, we use CPP macros in every module that declares an AST data type. We could perhaps put these all together in a single module and turn off warnings for orphan instances for that one module?


MatthewDaggitt:
Do you have any timings associated with this? How much does it speed it up? :smile:

wenkokke:
Makes a decent portion of the AST strict and pushes through ghc-debug and nothunks instrumentation, guarded by flags.

Please, can we merge this now? The merge conflicts are truly horrible.

wenkokke:
> Do you have any timings associated with this? How much does it speed it up? 😄

Haven't measured speedup yet, but our AST is still quite thunky, and a lot of problems can hide behind a single thunk.

wenkokke:
There's also a number of instance and pattern signatures I filled out as I went through. Sorry.

wenkokke:
The current error is:
```diff
    logic-and-comparisons
      TypeCheck:                OK (0.15s)
+      LossFunction:            OK (0.15s)
-      LossFunction:            FAIL (0.15s)
-        Missing output file logic-and-comparisons.loss.json
-        Contents of stderr differ:
-        0a1,2
-        > vehicle: Something went wrong internally. Please report the error shown below to `https://github.com/vehicle-lang/vehicle/issues`.
-        > Error: Unable to lower 'not' through gtRat
-        
-        Use -p '$0=="Tests.compile.logic-and-comparisons.LossFunction"' to rerun this test only.
      LossFunction-Godel:       OK (0.15s)
      LossFunction-Lukasiewicz: OK (0.22s)
      LossFunction-Yager:       OK (0.25s)
      LossFunction-Product:     OK (0.28s)
```
I really don't see how I could've messed with this directly, so I'm going to assume that this is a translation that relies on laziness?

vehicle/src/Vehicle/Compile.hs
```diff
@@ -197,6 +198,8 @@ parseProgText modul txt = do
     Left  err  -> throwError $ ParseError err
     Right prog -> case traverseDecls (parseDecl modul) prog of
       Left err    -> throwError $ ParseError err
+      -- If compiled with +ghc-debug, check that the elaborated program
+      -- does not contain any thunks.
```
MatthewDaggitt: What does this comment mean here? We're not doing anything special here?

vehicle-syntax/src/Vehicle/Syntax/AST/Binder.hs
```diff
@@ -1,14 +1,19 @@
+{-# LANGUAGE CPP #-}
```
MatthewDaggitt, line 1: Should we just be adding this option in the Cabal file? It's in a lot of modules...

cabal.project.ghc-debug
```diff
@@ -1,15 +1,23 @@
...
--- or run vehicle via `./scripts/vehicle-debug`.
+-- Use `cabal` with `--project-file=cabal.project.ghc-debug`
+-- or run vehicle via `./scripts/vehicle-ghc-debug`.
 
 import: cabal.project
 
+package vehicle-syntax
+  flags:
+    -- Necessary for Vehicle AST to implement NoThunks
+    +nothunks
```
MatthewDaggitt, line 11: Do we really want a separate flag and cabal project file for each debugging feature? I feel we should have a single `debug` flag and project file? Otherwise I can see it spiralling out of control as we add more debugging.

vehicle/tests/golden/NewTest.hs
```diff
@@ -1,5 +1,28 @@
+{-# LANGUAGE CPP #-}
+
+module Main where
+
 import System.Environment (getArgs)
 import Vehicle.Test.Golden.TestSpec.NewTestSpec (newTestSpec)
 
+#if ghcDebug
```
MatthewDaggitt, line 8: Why are we adding debug instrumentation to `NewTest`?

MatthewDaggitt (commented):
Thanks! I've added a few comments.

I've also pushed a commit reverting the syntax only changes. It's not that I don't agree with some of them, it's just that the code base is currently reasonably consistently formatted! Partially changing the formatting in some places in the middle of another PR is a tad sub-optimal and is going to result in a very messy style.

If you would like to change the formatting, then better to do it everywhere in a single dedicated PR :+1: 

**Edit:** Sorry, just saw your message on Slack

MatthewDaggitt:
Fix for the failing test incoming in https://github.com/vehicle-lang/vehicle/pull/329

cabal.project.ghc-debug
MatthewDaggitt: The answer is yes as we want to debug thunks vs test for no thunks separately.

vehicle/tests/golden/NewTest.hs
MatthewDaggitt: Decision: remove this

wenkokke: Removed.

cabal.project.ghc-debug
wenkokke: I'll leave these for now following our discussion.

PR #330, opened 2022-12-15
Switch to ormolu

wenkokke:
Advantages:

- more robust, since it handles multi-package repositories naively;
- more consistent, since it is more opinionated and leaves less room for us to format things manually;
- better diffs, since there are no cross dependencies such as those introduced by alignment;
- idempotent, since ormolu refuses to format a file if it detects non-idempotence;
- checked on CI with the ormolu action.


MatthewDaggitt:
I _really hate_ the loss of alignment, as I do think it makes it far far more readable. 

But I do recognise the other benefits so I'm very reluctantly merging it in.

PR #333, opened 2022-12-16
Make AST less strict for now

wenkokke:
Remove bang patterns for recursive arg, binder, and expr nodes, as well as lists and non-empty lists.


wenkokke:
Timing versus dev now.

PR #334, opened 2022-12-16
Fix tests

wenkokke:
This PR:
- restructures `vehicle.cabal`:
  + all flags;
  + common stanzas corresponding to flags;
  + `common-language`;
  + `common-library` for settings that apply to all libraries;
  + `common-executable` for settings that should apply to all executables.
- enables error-on-warnings for _all_ components, not just `vehicle:lib:vehicle`;
- fixes all resulting errors, including:
  + a seeming error in copying the needed files in the golden test framework?
  + an error in the unit tests for quantifier lifting (test alpha eq between the input term and itself, leave expected unused);
- disables unit test for quantifier lifting (all seem to have broken).

### DIFF
--- a/vehicle/app/Main.hs
+++ b/vehicle/app/Main.hs
@@ -4,7 +4,6 @@ module Main where
 
 import GHC.IO.Encoding (setLocaleEncoding, utf8)
 import Options.Applicative (execParser)
-import System.Environment (getArgs)
 import Vehicle (run)
 import Vehicle.CommandLine (commandLineOptionsParserInfo)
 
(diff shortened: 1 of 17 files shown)


MatthewDaggitt:
Hang on, why have you disabled the quantifier lifting tests? These weren't broken on `dev`...

wenkokke:
> Hang on, why have you disabled the quantifier lifting tests? These weren't broken on `dev`...

If you check the diff, you'll see:
Screenshot 2022-12-17 at 17 16 33
Which "fixes" the test, but also breaks them, if that makes sense.

Issue #331, opened 2022-12-16
275% performance drop


MatthewDaggitt:
The recent commit 5386b68fe3026362f1ea22035db7dc3e88471f9b introducing strictness and no thunks increased the time required to run the test suite from ~2 seconds on my high-specs laptop to 7.5 seconds. This is moderately disastrous as it can't easily be backed out again without also backing out [ca99b8d](https://github.com/vehicle-lang/vehicle/commit/ca99b8da9e5aabde2c94b758bb4141fbe53ebed5).

As I think I mentioned in the chat before, we really shouldn't be making any performance changes that aren't clearly cutting out code execution without a working benchmark suite.

wenkokke:
We can simply remove the bangs and leave the other changes intact.

wenkokke:
I'm having trouble finding a performance regression. If anything, there's a slight improvement.

## benchmark 78508913bd1104468dd161db2729f6e5739f2472

```
Benchmark 1: cabal test all
  Time (mean ± σ):     19.570 s ±  0.129 s    [User: 16.862 s, System: 3.403 s]
  Range (min … max):   19.333 s … 19.766 s    10 runs
```

## benchmark `benchmark-before`

Includes #330, #332, and #334.

```
Benchmark 1: cabal test all
  Time (mean ± σ):     20.030 s ±  0.118 s    [User: 16.982 s, System: 3.342 s]
  Range (min … max):   19.917 s … 20.301 s    10 runs
```

## benchmark `benchmark-before-hasmap`

Includes #330 and #334.

```
Benchmark 1: cabal test all
  Time (mean ± σ):     20.043 s ±  0.100 s    [User: 16.967 s, System: 3.324 s]
  Range (min … max):   19.955 s … 20.240 s    10 runs
```

## benchmark `benchmark-after`

Includes #315, #330,  #334.

```
Benchmark 1: cabal test all
  Time (mean ± σ):     19.298 s ±  0.082 s    [User: 16.361 s, System: 3.333 s]
  Range (min … max):   19.174 s … 19.440 s    10 runs
```

wenkokke:
I'm not sure if it's strictness & nothunks, but there's definitely a performance hit in the past few commits:

### benchmark 78508913bd1104468dd161db2729f6e5739f2472
```
Benchmark 1: cabal test all
  Time (mean ± σ):      9.671 s ±  0.056 s    [User: 27.081 s, System: 4.919 s]
[...]

PR #346, opened 2022-12-31
Fix performance problems by temporarily changing cabal library used for executable component

MatthewDaggitt:
Fixes the latest performance problems reported in https://github.com/vehicle-lang/vehicle/issues/331. Reduces benchmark execution time on my machine from `27.984 s ±  2.241 s` to `5.730 s ±  0.218 s` on my machine. And from 22 s to 2 s when running the test suite normally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,20 +41,20 @@ jobs:
         cabal-extra-args: [""]
         include:
           # Build with -fghc-debug:
-          - os_and_name: [ubuntu-latest, Linux]
-            ghc-version: "9.2.4"
-            cabal-version: "3.8"
-            cabal-extra-args: "-fghc-debug"
+          #- os_and_name: [ubuntu-latest, Linux]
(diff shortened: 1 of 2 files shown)


MatthewDaggitt:
@wenkokke the `GHC 9.2.4 -fghc-debug` and `GHC 9.2.4 -fghc-debug -fnothunks` CI tests are taking over an hour, is that expected?

MatthewDaggitt:
I've disabled the `-fghc-debug` builds as they seem to be never terminating. Unsure whether that's because of the change I've made to the `.cabal` file or not.

wenkokke:
> I've disabled the `-fghc-debug` builds as they seem to be never terminating. Unsure whether that's because of the change I've made to the `.cabal` file or not.

They're not running because ghc-debug requires the `-threaded` RTS option, and removing it basically deadlocks the process.

wenkokke:
This actually introduces a 100% performance penalty for me, going from 10s to 20s.

PR #348, opened 2022-12-31
Fix: only pass `-threaded` and `-N` if `--ghc-debug` is set

wenkokke:
@MatthewDaggitt please benchmark these changes, as the 2nd performance regression isn't detectable on my system.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,21 +40,27 @@ jobs:
         cabal-version: ["3.8"]
         cabal-extra-args: [""]
         include:
-          # Build with -fghc-debug:
-          - os_and_name: [ubuntu-latest, Linux]
-            ghc-version: "9.2.4"
-            cabal-version: "3.8"
-            cabal-extra-args: "-fghc-debug"
           # Build with -fnothunks:
           - os_and_name: [ubuntu-latest, Linux]
             ghc-version: "9.2.4"
             cabal-version: "3.8"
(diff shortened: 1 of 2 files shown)


wenkokke:
This should fix performance _for you_ but actually still introduces a 100% slowdown for me, going from 10s to 20s.

MatthewDaggitt:
Hmmm so
- current dev: benchmark `5.911 s ±  0.196 s`, repeated manual individual runs `1.96s`
- this branch: benchmark `8.871 s ±  0.027 s`, repeated manual individual runs `9.73s`

So there's still some slow-down....

wenkokke:
How, though? There literally shouldn't be any difference if you're not building with ghc-debug?

Could it be noise? Are you running these benchmarks using scripts/benchmark on a computer with no other programs running?

MatthewDaggitt:
> How, though? There literally shouldn't be any difference if you're not building without ghc-debug?

Did you mean this double-negative?

> Could it be noise? Are you running these benchmarks using scripts/benchmark on a computer with no other programs running?

Yup. No other programs running. The first numbers are obtained by running `./scripts/benchmark`.

Are you free to call at some point tomorrow and we can try and work it out?

wenkokke:
> > How, though? There literally shouldn't be any difference if you're not building without ghc-debug?
> 
> Did you mean this double-negative?

I did not.

wenkokke:
Maybe it’s time for the Haskell mailing list?

Issue #390, opened 2023-01-27
Unable to find consistent type in mnist spec


ndslusarz:
Trying to compile modified mnist spec:

```
@network
mnist : Tensor Rat [28, 28] -> Vector Rat 10

--------------------------------------------------------------------------------
-- Definition of robustness around a point

-- First we define the parameter `epsilon` that will represent the radius of the
-- ball that we want the network to be robust in. Note that we declare this as
-- a parameter which allows the value of `epsilon` to be specified at compile
-- time rather than be fixed in the specification.

--epsilon : Rat
--epsilon = 0.01

--delta : Rat
--delta = 0.02

-- Next we define what it means for an image `x` to be in a ball of
-- size epsilon around 0.
boundedByEpsilon : Tensor Rat [28, 28] -> Bool
boundedByEpsilon x = forall i j . -0.01 <= x ! i ! j <= 0.01

--The disjunction avoids having to define absolute value
boundedByDelta : Tensor Rat [28, 28] -> Tensor Rat [28, 28] -> Bool
boundedByDelta x y = forall i . (-0.02 <= (((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02) 
                                or (-0.02 <= (-((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02)
                                or (-0.02 <= (((mnist x) ! i) + ((mnist y) ! i )) <= 0.02)
                                or (-0.02 <= (-((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02)


robustSingleInput : Tensor Rat [28, 28] -> Bool
robustSingleInput image = forall pertubation .
    let perturbedImage = image - pertubation in
    boundedByEpsilon pertubation => True

@property
robust1 : Bool
robust1 = forall x . robustSingleInput x
```

 spec throws an error where it's unable to find a type for `perturbation`

```
"Error during specification compilation: " + result.stderr.decode("UTF-8")  
Exception: Error during specification compilation: Error at Line 44, Columns 32-33: unable to find a consistent type for the overloaded expression '-'. In particular 28 != image.
Fix: check your types
```

MatthewDaggitt:
I'm afraid I don't get this error on `dev` branch at all. Do you get on `dev`? If not, then I'm afraid that means its a change on your personal branch that is responsible.

ndslusarz:
[...]

Issue #386, opened 2023-01-23
unexpected expression '(Vector Rat) 28' - compilation to LossFunctions


ndslusarz:
When trying to compile to loss functions a modified mnists-spec (slightly simplifed to avoid other issues and with a different robustness not referring to label but rather difference in value between outputs) which can be found below:

```
-- Inputs and outputs

-- Define the type for our input images
--type Image = Tensor Rat [28, 28]

--------------------------------------------------------------------------------
-- Network

-- Declare the network used to classify images. The output of the network is a
-- score for each of the digits 0 to 9.
@network
mnist : Tensor Rat [28, 28] -> Vector Rat 10

--------------------------------------------------------------------------------
-- Definition of robustness around a point

-- First we define the parameter `epsilon` that will represent the radius of the
-- ball that we want the network to be robust in. Note that we declare this as
-- a parameter which allows the value of `epsilon` to be specified at compile
-- time rather than be fixed in the specification.

--epsilon : Rat
--epsilon = 0.01

--delta : Rat
--delta = 0.02

-- Next we define what it means for an image `x` to be in a ball of
-- size epsilon around 0.
boundedByEpsilon : Tensor Rat [28, 28] -> Bool
boundedByEpsilon x = forall i j . -0.01 <= x ! i ! j <= 0.01

--The disjunction avoids having to define absolute value
boundedByDelta : Tensor Rat [28, 28] -> Tensor Rat [28, 28] -> Bool
boundedByDelta x y = forall i . (-0.02 <= (((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02) 
                                or (-0.02 <= (-((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02)
                                or (-0.02 <= (((mnist x) ! i) + ((mnist y) ! i )) <= 0.02)
                                or (-0.02 <= (-((mnist x) ! i ) - ((mnist y) ! i )) <= 0.02)


robustSingleInput : Tensor Rat [28, 28] -> Bool
robustSingleInput image = forall pertubation .
    let perturbedImage = image - pertubation in
    boundedByEpsilon pertubation => boundedByDelta image perturbedImage

@property
robust1 : Bool
robust1 = forall x . robustSingleInput x

```
I get an error `encountered unexpected expression '(Vector Rat) 28' during compilation to loss functions.`

[...]

PR #395, opened 2023-01-30
Loss translation errors fix

ndslusarz:
This has the changes fixing Issues #390 and #386 - fixing Subtraction and Addition in loss function translation so that it now works properly for Vectors and a minor change to make the python script Windows compatible.


ndslusarz:
Actually I did some further testing - the small edit to make it compatible with Windows doesn't work on Windows:

```
command = " ".join([vehicle] + args)
    result = subprocess.run(command, capture_output=True, shell=True)
```

Adding `shell=True` does make it Windows friendly however the addition of `" ".join` to pass the command as string that was necessary for Linux (once the shell was set to True) once again breaks the Windows side (while the combination of `shell=True` with `command=[vehicle]+args` does work) with error:

```
Exception: Problem during compilation: The syntax of the command is incorrect.
Command was: C:\cabal\bin\vehicle.EXE compile --target LossFunction --specification vehicle-python/tests/mnist.vcl --outputFile C:\Users\MALCAD~1\AppData\Local\Temp\tmpaub_kpxhloss_function.json --network mnist:<keras.engine.sequential.Sequential object at 0x0000011537C32128>
```

Not too much of a priority I'd say, considering other issues with loss translation, since the fix is easy to do locally, but something I'll look at when I have some more time.

MatthewDaggitt:
Okay, I've removed the command line changes and the dataset and will just merge in the loss function translation changes.

PR #406, opened 2023-02-03
Change Python loss function translation to use exec

### LINKED ISSUE #403: Loss functions - problem with quantifier compilation (comparison between 'function' and 'float')
When trying to run the python script for the modified mnist example (both can be found on `loss_experiments `branch in `vehicle-python/tests` as `mnist.vcl` and `test_mnist.py`) it returnes an error:

```
File "e:\vehicle\vehicle\vehicle-python\tests\__init__.py", line 325, in result_func        
    max_loss = max(max_loss, body_loss(context))
TypeError: '>' not supported between instances of 'function' and 'float'
```

This errors out inside the compilation of a forall quantifier. Speicifcally it seems that `body_loss(context)` which is supposed to compile the value of everything inside the forall, does not actually compile to a number and is still a function. As in specification a helper function is called immeditaely after forall I suspect there is something wrong in how the translation handles translating these, probably linked to the latest fix which makes the property compile first. I might be wrong as I haven't been able to fix it yet.

MatthewDaggitt:


wenkokke:
Oh dear, is there any way we can avoid this?

wenkokke:
This was also merged with failing checks, specifically, mypy complains about the Python code.

Should we roll this PR back for now?

MatthewDaggitt:
> Oh dear, is there any way we can avoid this?

Having implemented it the "right" way first, I think there are _significant_ advantages to this approach. Namely:

1) We can actually see the string representation of the function for debugging purposes (which is massive).
2) It's significantly faster as we translate the compiled function to byte-code directly, and we eliminate the incredibly deep nested recursion.

I also restrict the global and local scope explicitly, and all variable names have to go through the very restrictive Vehicle parser first. I think it's highly unlikely there's valid Vehicle programs that are also injection attacks in Python.

> This was also merged with failing checks, specifically, mypy complains about the Python code.
>
> Should we roll this PR back for now?

No. I merged this in overriding the checks because Natalia urgently needed this code to run her experiments. Furthermore, the linter failed because the code wasn't all typed. It wasn't typed before, and it is partially typed now so it is better than it was before.

PR #409, opened 2023-02-03
Fix type error introduced by #406

wenkokke:


MatthewDaggitt:
Ah sorry, I misunderstood the error message. Thought it was complaining that I hadn't provided a type, not that it was the wrong type.

Thanks for the fix!

PR #445, opened 2023-03-09
Overhaul of verify mode and added typeCheck mode

MatthewDaggitt:
See CHANGELOG entries for description


wenkokke:
I'd ask to call the command `check` for brevity?

Isn't typeCheck mode already exposed via `compile -t TypeCheck`?

MatthewDaggitt:
> I'd ask to call the command check for brevity?

The mode `check` was already reserved for checking the proof-cache, but in order to follow your suggestion I have renamed that command `validate`, and changed `typeCheck` to `check`.

> Isn't typeCheck mode already exposed via compile -t TypeCheck?

Yes, but semantically this makes far more sense as now the mode `compile` and its target option `-t` really is reserved for compilation targets.

PR #459, opened 2023-03-23
Fix `alex` version

MatthewDaggitt:
Trying to work out what's causing the CI to fall over in e.g. https://github.com/vehicle-lang/vehicle/actions/runs/4498380696/jobs/7914938112

### DIFF
--- a/vehicle-syntax/vehicle-syntax.cabal
+++ b/vehicle-syntax/vehicle-syntax.cabal
@@ -156,6 +156,9 @@ library
     , these              >=1.0    && <2
 
   build-tool-depends:
-    , alex:alex    >=3
+    , alex:alex    ==3.2.7
     , BNFC:bnfc    >=2.9.4
     , happy:happy  >=1.19
+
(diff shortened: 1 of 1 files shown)


wenkokke:
We may need to report this upstream?

MatthewDaggitt:
To the `setup-haskell` repo or to `alex` directly?

MatthewDaggitt:
I'll try merging this in now and seeing if it'll fix it.

wenkokke:
To haskell/cabal.

This is the error we’re running into:
https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/Simple/Program/Db.hs#L489-L492

Which means that the function finding the Alex version in this definition is failing:
https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/Cabal/src/Distribution/Simple/Program/Builtin.hs#L239-L247

MatthewDaggitt:
Drat the error has reoccured.

Opened https://github.com/haskell/cabal/issues/8876

PR #463, opened 2023-03-24
Loss experiments

ndslusarz:
Bit of a large pull request but all confined to loss functions.

On the haskell side STL translation has been added (not handled by python fully yet) and some bugs in the interpretation of comparisons were fixed.

On the python side there is experiments for mnist (including test file, constraint accuracy method and a folder with separate setup for a series of experiments run by script) together with some fixes to the translation translation of loss function in general.


MatthewDaggitt:
Thanks! Will review this first thing on Monday :+1:

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -51,11 +53,12 @@ compile resources logic typedProg = do
   let unnormalisedProg = fmap unnormalised expandedProg
   progWithoutInstanceArgs <- resolveInstanceArguments unnormalisedProg
   let descopedProg = descopeNamed progWithoutInstanceArgs
-  possiblyNotNormalisedProg <- case compileNot (implementationOf logic) of
-    Just {} -> return descopedProg
-    Nothing -> lowerNots logic descopedProg
+  reformattedProg <- reformatLogicalOperators logic descopedProg --case compileNot (implementationOf logic) of
+    --Just {} -> return descopedProg
+    --Nothing -> reformatLogicalOperators logic descopedProg
```
MatthewDaggitt: Remove comments

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -179,8 +196,15 @@ compileBuiltinFunction f t args = case f of
...
-  V.Or -> compileOp2 (compileOr t) t args
+  V.And  ->  case compileAnd t of
+      Left binaryAnd -> compileOp2 binaryAnd t args
+      Right naryAnd -> return (naryAnd args)
+  V.Or  ->  case compileOr t of
+      Left binaryOr -> compileOp2 binaryOr t args
+      Right naryOr -> return (naryOr args)
+
+  --V.And -> compileOp2 (compileAnd t) t args
+  --V.Or -> compileOp2 (compileOr t) t args
```
MatthewDaggitt: Remove commented out code

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -108,6 +113,18 @@ type MonadCompileLoss m =
...
+flattenAnds arg = case argExpr arg of
+  V.AndExpr _ [e1, e2] -> flattenAnds e1 <> flattenAnds e2
+  _ -> [arg]
+
+flattenOrs :: InputArg -> NonEmpty InputArg
+flattenOrs arg = case argExpr arg of
+  V.OrExpr _ [e1, e2] -> flattenOrs e1 <> flattenOrs e2
+  _ -> [arg]
+--addToCtx :: MonadCompileLoss m => InputBinder -> m a -> m a
+--addToCtx binder = local (\(a, b, c, ctx) -> (a, b, c, V.nameOf binder : ctx))
```
MatthewDaggitt: Remove commented out code.

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
```diff
@@ -108,6 +113,18 @@ type MonadCompileLoss m =
     MonadReader (V.NetworkContext, DifferentiableLogic, V.DeclProvenance) m
   )
 
+flattenAnds :: InputArg -> NonEmpty InputArg
+flattenAnds arg = case argExpr arg of
+  V.AndExpr _ [e1, e2] -> flattenAnds e1 <> flattenAnds e2
+  _ -> [arg]
```
MatthewDaggitt: It might be tidier to make `flattenAnds` and `flattenOrs` sub-functions of `reformatLogicalOperators` in the same way that `lowerNot` is.

yager_training/keras_metadata.pb
MatthewDaggitt, line 1: We should not be including these trained models in the commits, unless they are specifically used as inputs (i.e. rather than outputs) for testing purposes.

vehicle/tests/golden/compile/test-mnist/test-mnist.vcl
```diff
@@ -0,0 +1,46 @@
...
+--------------------------------------------------------------------------------
+-- Definition of robustness around a point
+
+-- First we define the parameter `epsilon` that will represent the radius of the
+-- ball that we want the network to be robust in. Note that we declare this as
+-- a parameter which allows the value of `epsilon` to be specified at compile
+-- time rather than be fixed in the specification.
+
+--epsilon : Rat
+--epsilon = 0.01
```
MatthewDaggitt: Please could you open issues for any bugs you've found in the translation of loss functions so we can track them?

vehicle-python/vehicle/command_line.py
```diff
@@ -37,7 +42,12 @@ def call_vehicle_to_generate_loss_json(
         args = [
             "compile",
             "--target",
+            # use the string below to specify which translation is used. the options are:
+            # LossFunction (uses DL2 by default), LossFunction-DL2, LossFunction-Godel, LossFunction-Lukasiewicz,
+            #  LossFunction-Product, LossFunction-STL (in progress, not currently working)
             "LossFunction",
+            # for experiments,
+            # os.getenv('DLV'),
```
MatthewDaggitt: What does this commented out line do? Should it be removed?

vehicle-python/vehicle/__init__.py
```diff
@@ -228,19 +249,19 @@ def _translate_quantifier(self, contents: Dict[Any, Any]) -> str:
...
-        # max_loss = np.NINF
-        # min_loss = np.Inf
         # We generate 10 samples, have to change it in the future
         if quantifier == "All":
             op = "max"
         elif quantifier == "Any":
             op = "min"
+        # elif quantifier == "ForallIn":
+        #    op = "max"
+        #    n =
```
MatthewDaggitt: Remove commented out code?

vehicle-python/vehicle/__init__.py
```diff
@@ -54,7 +53,7 @@ def __init__(
         self.quantifier_sampling: Dict[str, Callable] = quantifier_sampling
         self.current_decl = None
 
-        self.debug: bool = False
+        self.debug: bool = True
```
MatthewDaggitt: Please could you turn `debug` back to `False` so we don't get lots of output in the tests?

vehicle-python/tests/test_LossFunctionTranslation.py
```diff
@@ -6,6 +6,12 @@
 import tensorflow as tf
 import tensorflow.keras as keras
 
+if __name__ == "__main__" and __package__ is None:
+    from os.path import dirname as dir
+    from sys import path
+
+    path.append(dir(path[0]))
```
MatthewDaggitt, line 13: This particular 5 lines occurs rather a lot? What is the purpose?

vehicle-python/tests/mnist_benchmark_setup/vehicle.py
```diff
@@ -0,0 +1,278 @@
+import typing
+from typing import Any, Callable, Dict, Optional, Set, Tuple
+
+import tensorflow as tf
+from command_line_experiment_setup import call_vehicle_to_generate_loss_json
+from utils import internal_error_msg
+
+
+def generate_loss_function(
```
MatthewDaggitt: These seems to be an entirely copy of the main translation file? Surely we don't want to commit it twice?

MatthewDaggitt (commented):
Okay my main comment is that a lot of these look like your experiments and therefore shouldn't be committed to the main Vehicle repository. In general the main repo should only contain:

1. tests - should be able to be run quickly automatically by the automatic CI integration suit.
2. worked example - should be well-documented code with copious explanation of what's going on.

I suggest that for this PR, you remove all the experiment-specific code, and simply retain the changes to the main Vehicle code base. You can then submit another PR with any code that you think should be added either as a test or a worked example?

vehicle-python/tests/test_LossFunctionTranslation.py
ndslusarz: The way we import translation is from a sibling folder `vehicle` that houses `__init__` file together with `utils` and `command_line`. The problem is while I'm not sure if it worked on Linux, haven't been able to test that, I can confirm import like that does not work on both Mac and Windows. 
This is a workaround that makes it work and allows to keep the main impementation separate from tests in a sibling forlder - essentially adding the path to the folder to places where python will look for modules to import from. It avoids a situation where you have to move the files to the same folder if you want to run it on a different system.

vehicle-python/vehicle/command_line.py
ndslusarz: That was for experiment setup where different type of loss function was passed to command line. Will remove from here, not needed in general.

vehicle/tests/golden/compile/test-mnist/test-mnist.vcl
ndslusarz: Will do, forgot about this issue, sorry.

yager_training/keras_metadata.pb
ndslusarz: Sorry, missed it when deleting others, will get rid of it.

vehicle-python/tests/mnist_benchmark_setup/vehicle.py
ndslusarz: This had to be copied since it uses a slightly modified version of command_line.py for the benchmark experiments - I will delete those and keep them local anyway as per your other comment, I'll leave the singular example worked out only (not the setup to run multiples)

vehicle/src/Vehicle/Backend/LossFunction/Compile.hs
ndslusarz: Will do.

yager_training/keras_metadata.pb
ndslusarz: Sorry, must have missed it when deleting others, will delete.

ndslusarz:
Alright, will do - corrected most of the things you pointed out, deleted the full setup for experiments (the mnist example is still there for now, I will clean that up tomorrow - probably comment it well inside etc. since without it there will be no working exampe of loss functions). Still need to open that forgotten issue so I left that thread open to remember.

ndslusarz:
Did all of the changes aside from removing mnist example - not sure if you'd prefer that example be removed from repo for now, added as a test in a python folder (that's how what state it is now), be wrapped in a neat folder as a more proper worked out example...?

vehicle-python/vehicle/__init__.py
```diff
@@ -206,7 +227,7 @@ def _translate_at(self, contents: Dict[Any, Any]) -> str:
     def _translate_network(self, contents: Dict[Any, Any]) -> str:
         model = contents[0]
         input_losses = ", ".join([self._translate_expression(c) for c in contents[1]])
-        return f"{model}({input_losses}, training=True)"
+        return f"{model}(expand({input_losses}, axis=0), training=True)[0]"
```
MatthewDaggitt, line 232: Could you add a comment here explaining why there is an `expand` inserted here?

vehicle-python/vehicle/__init__.py
```diff
@@ -168,6 +170,25 @@ def _translate_maximum(self, contents: Dict[Any, Any]) -> str:
         loss_2 = self._translate_expression(contents[1])
         return f"max({loss_1}, {loss_2})"
 
+    # NOT WORKING CURRENTLY
+    def _translate_exponential_and(self, contents: Dict[Any, Any]) -> str:
+        # translation of conjunction solely for the STL based translation
+        conjuncts = [self._translate_expression(c) for c in contents]
+        a_min = min(conjuncts)
+        v = 3  # a constant to be set by user (refer to Varnai and Dimarogonas, "On Robustness Metrics for Learning STL Tasks." 2020)
```
MatthewDaggitt: Could comments go on the line above rather than the same line? It would also be nice to keep the width of the comment shorter by splitting it over multiple lines.

vehicle-python/vehicle/__init__.py
```diff
@@ -168,6 +170,25 @@ def _translate_maximum(self, contents: Dict[Any, Any]) -> str:
         loss_2 = self._translate_expression(contents[1])
         return f"max({loss_1}, {loss_2})"
 
+    # NOT WORKING CURRENTLY
```
MatthewDaggitt, line 173: If this isn't working currently, wouldn't it be better to comment this code out and throw an error? Rather than lulling the user into thinking it's working?

vehicle-python/tests/test_LossFunctionTranslation.py
MatthewDaggitt: Okay, sure. I guess we'll revisit it when we make an actual Python package.

MatthewDaggitt (commented):
> Did all of the changes aside from removing mnist example - not sure if you'd prefer that example be removed from repo for now, added as a test in a python folder (that's how what state it is now), be wrapped in a neat folder as a more proper worked out example...?

The latter sounds great, thanks!

Also before you commit again, do you think you could follow the [instructions](https://github.com/vehicle-lang/vehicle/blob/dev/HACKING.md#pre-commit-hooks) in `HACKING.md` to install the pre-commit hooks? That should auto format your code on commit.

vehicle-python/vehicle/__init__.py
ndslusarz: I genuinely cannot remember - I believe this was related to the shape of input to the NN which we had issues with but I don't want to add incorrect information here.

MatthewDaggitt: Okay... In future when you add code like this that is likely to be mysterious in future, please can you add explanatory comments about it at the time! At the very least explaining what breaks if you remove it.

MatthewDaggitt:
Okay I've tried my best to fix all the tests.

However when trying to generate loss functions for `test-mnist.vcl` I'm getting the error:
> encountered unexpected expression ''forallIndex' is not handled at the moment for loss function translation.' during compilation to loss functions.

I'm a bit confused. It's usually good practice to try and run the tests you added :wink: 

Should this work or is it another bug?

ndslusarz:
Do you mean the test_mnist_custom_loss.py? It runs without any errors for me. The mnist.vcl spec for that example doesn't have forallIn in it (I'm not sure which file you mean by test-mnist.vcl).

When it comes to forallIn producing an error message that is correct, it's not handled yet, I think we discussed it at some point (I can add an issue for implementing it, I thought it was already there but can't find it now).

ndslusarz:
Sorry, had to repost, couldn't find the correct test-mnist.vcl before. When it comes to forallIn producing an error message that is correct, it's not handled yet, I think we discussed it at some point (I can add an issue for implementing it, I thought it was already there but can't find it now). 

I think that test is unneccessary actually, it seems to be an imperfect copy of the mnist.vcl in examples (but with forallIn)... I must have commited it by accident, I kept it for local experiments only when fixing some things with Compile a while back so I didn't pay attention to it failing and didn't notice it got commited. I'll get it removed.

MatthewDaggitt:
Thanks, merging in :+1:

PR #482, opened 2023-04-03
Build wheels for Vehicle

wenkokke:
This PR:

- adds the project `py-vehicle-lang` which builds Python bindings for the Vehicle CLI;
- renames the project `vehicle-python` to `py-vehicle-lang-tensorflow`;
- adds support for GHC 9.6.1 on CI & fixes various bugs relating to GHC 9.6.1 in the `vehicle` and `vehicle-syntax` projects;
- changes the preferred GHC version to GHC 9.4.4;
- bumps the minimum Cabal version to 3.8;


pbcopy
```diff
@@ -0,0 +1,6 @@
+  system
+  3.7.16
+  3.8.16
+  3.9.16
+  3.10.10
+* 3.11.2 (set by PYENV_VERSION environment variable)
```
MatthewDaggitt: Was this file meant to be commited? It doesn't look machine parsable...

MatthewDaggitt (commented):
Super impressive. My only main query is the renaming of the main python project code to `py-vehicle-lang-tensorflow`. Only the code within the `__init__.py` is tensorflow-specific and within that only a tiny fragment of the actual code. I was kind of hoping got avoid having a whole separate Python project for each training framework...

wenkokke:
I think that's something we can work out later? It needs to be a separate package for now, but it's not being published, so we can refactor it later?

pbcopy
wenkokke: That should be removed.

PR #487, opened 2023-04-04
Fix #486 and #485

### LINKED ISSUE #485: Fix `--version` commandline option
Returns the printed Haskell structure

### LINKED ISSUE #486: Move `defaultMain` logic into `Vehicle.hs`
Expose a method that is of type `[String] -> IO Int`. Catch any exit codes and pass them on.

wenkokke:
This PR fixes #486 by:
- moving the `defaultMain` function to `Vehicle.hs` (renamed to `main`);
- adding options for redirecting the standard output and error streams in Vehicle main;
- creating a corresponding function of type `[String] -> IO Int` which catches any `ExitCode` and converts it to an int.

This PR also fixes #485 by using `showVersion`.

Furthermore, it:
- updates the Python bindings so that the exit code is sent to Python; and
- exposes the `vehicleSync` function from `vehicle-lang` which appropriately redirects and grabs the output streams.

Breaking changes:
- `--redirectLogs` changed to `--redirect-logs`


MatthewDaggitt:
In future, can we make sure to add any breaking changes to the `Changelog.md`? In this case the `--redirectlogs` renaming?

PR #518, opened 2023-04-25
Use file-embed to embed standard library

### LINKED ISSUE #517: Use file-embed to embed standard library
See [file-embed](https://hackage.haskell.org/package/file-embed).

wenkokke:
Fix #517.

Changes the std version to match the vehicle version by default.

A future improvement would be to compile std at Vehicle compile time and embed the vclo file as well.


MatthewDaggitt:
Thanks looks great!

PR #519, opened 2023-04-28
Implemented object file versioning and general versioning improvements

MatthewDaggitt:
As discussed added a build flag `release-build` to the cabal file which can be used to turn off the addition of additional information to the version.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -30,6 +30,9 @@
 
 * Fixed erroneous evaluation of when dividing two rationals together.
 
+* Fixed bug where the compiler would sometimes hang when reading `.vclo` files created with an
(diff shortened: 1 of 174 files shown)


MatthewDaggitt:
Looks like the CI isn't quite right for `py-vehicle-lang-tensorflow`? It's running before `py-vehicle-lang` is being built? And therefore can't find the new versions?

wenkokke:
I made the change we discussed, where it uses the published version only. Therefore, the version vehicle-lang-tensorflow relies on shouldn't be updated until we actually publish a new version of vehicle.

MatthewDaggitt:
Thanks!

PR #521, opened 2023-05-01
Preparations for implementing @noinline in NBE

MatthewDaggitt:
Turns out implementing @noinline requires some medium size changes to how we handle external resource inlining. This PR adds the `@noinline` syntax to the front-end makes those changes in preparation for the actual implementation.

 In particular before, we were sneakily renormalising the entire expressions in the `expandResources` pass. This has been made much more explicit by actually forcing the `compileQueries` code to do the normalisation itself. This also makes it clear that we never actually use the glued expression representation outside of type-checking itself....

### DIFF
--- a/vehicle-syntax/src/Vehicle/Syntax/AST/Decl.hs
+++ b/vehicle-syntax/src/Vehicle/Syntax/AST/Decl.hs
@@ -53,6 +53,11 @@ bodyOf = \case
   DefFunction _ _ _ _ e -> Just e
   DefAbstract {} -> Nothing
 
+annotationsOf :: GenericDecl expr -> [Annotation]
+annotationsOf = \case
+  DefFunction _ _ anns _ e -> anns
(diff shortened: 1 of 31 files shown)


wenkokke:
I feel the purpose of the glued representation is to allow us to reuse the normalised expression after type checking, without having to start from scratch. 

What's currently preventing that from happening?

MatthewDaggitt:
> What's currently preventing that from happening?

It's our implicit invariant that we want to avoid renormalising (which we already break in the query compilation). We get new information after-type checking in the form of the parameter and dataset values. We don't know this at type-checking time, and so if we want to preserve the invariant of being weakly normalised we need to _renormalise_. Will discuss in our next call?

MatthewDaggitt:
I honestly have no idea what's going on with this one. Non deterministic test failures galore. None of which look like this PR's fault....

PR #526, opened 2023-05-04
Replace `normApp` with smart `App` synonym

wenkokke:
This PR removes `normApp` and `renormArgs` in favour of an `App` constructor that maintains the invariant.


MatthewDaggitt:
Looks great!

Issue #537, opened 2023-05-18
Release version of Vehicle versioning not working


MatthewDaggitt:
I ran:
```
pip install vehicle-lang
vehicle --version
```
and got
```
0.3.0+b11752b
```

This suggests the cabal `release-build` flag is not being set properly?

wenkokke:
It's not, because there's no release workflow on GitHub at the present.

MatthewDaggitt:
Ah right, that explains it! Which script does the distribution to `PyPy`?

wenkokke:
> Ah right, that explains it! Which script does the distribution to `PyPy`?

Me.

wenkokke:
Trusted publishers from GitHub now work, so I'll actually write a script now.

wenkokke:
Fixed.

PR #541, opened 2023-05-18
Automatically publish releases & ensure Vehicle Python is always built as a release build

wenkokke:


wenkokke:
Fixes #537, but only by _always_ building a release version for Python on CI.

PR #547, opened 2023-05-25
Prune type aliases

wenkokke:
This is part of a refactoring to prune the _many_ type aliases for the AST that exist within the codebase:

- Rename DBIndex -> Ix and DBLevel -> Lv
- Rename NormExpr -> Value
- Inline `Vehicle.Expr.DBExpr`
- Inline `Vehicle.Syntax.AST.InputExpr`
- Inline `Vehicle.Compile.Prelude.NamedExpr`
- Inline `Vehicle.Backend.LossFunction.Compile.InputExpr`
- Inline `Vehicle.Backend.Agda.Compile.OutputExpr`
- Inline `Vehicle.Compile.Scope.UnscopedExpr`
- Inline `Vehicle.Compile.Scope.ScopedExpr`
- Inline `Vehicle.Compile.Type.Core.UncheckedExpr`
- Inline `Vehicle.Compile.Type.Core.CheckedExpr`

I believe this PR should be merged ASAP, due to the high risk of conflicts.

Pruning of `NormalisableExpr` and the derived aliases requires a refactoring of the builtin types.

### DIFF
--- a/vehicle-syntax/src/Vehicle/Syntax/AST/Expr.hs
+++ b/vehicle-syntax/src/Vehicle/Syntax/AST/Expr.hs
@@ -21,15 +21,6 @@ module Vehicle.Syntax.AST.Expr
     UniverseLevel (..),
     Telescope,
 
-    -- * Input expressions
-    InputBinding,
-    InputVar,
-    InputArg,
-    InputBinder,
-    InputExpr,
-    InputDecl,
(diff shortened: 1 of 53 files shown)


wenkokke:
@MatthewDaggitt Is there any variant of the AST that doesn't pass `()` to the binder? I don't think I've found any.

PR #561, opened 2023-06-03
Convert test_pygments from Shell to Python

wenkokke:


wenkokke:
For some reason, it seems that the BNFC-generated lexer isn't built on Windows?

PR #565, opened 2023-06-05
Remove binder representation

wenkokke:


MatthewDaggitt:
Finally remembered what this was originally for! co-DeBruijn indices!

PR #570, opened 2023-06-09
Add support for TensorFlow

wenkokke:


wenkokke:
When this is merged, we can remove `py-vehicle-tensorflow` and potentially rename `py-vehicle-lang` to `vehicle-python`.

MatthewDaggitt:
> When this is merged, we can remove py-vehicle-tensorflow and potentially rename py-vehicle-lang to vehicle-python.

I'd be in favour of that :+1:

PR #597, opened 2023-06-24
Eager solving of tensor level variables during query compilation

### LINKED ISSUE #479: Improve performance of normalisation (again!)
Can never have too much performance. 

So currently it takes about 1.5 seconds to compile the Marabou queries per image in the mnist-dataset. The problem is that we first normalise all the vector quantifiers to individual quantified rationals, then convert to DNF (which blows up the number of queries by a factor of 10, i.e. the number of outputs) and then solve the resulting 784 simultaneous equations for the user-variables. This results in us resolving the same set of 784 simultaneous equations 10 times per image.

However, in theory this can be improved in two ways:
 - The easiest way is that we could cache the simultaneous equations and their solutions. This would mean that we would only solve the 784 simultaneous equations 1 time instead of 10 times.
 - The hard way would be to try and solve the simultaneous equations at the vector level and only then fall back to solving them after quantifier normalisation if that fails (and it may fail as we can't normalise higher order functions such as map and fold over tensor variables). Unfortunately that raises all sorts of hairy issues about normalisation, and vector level operations etc.

MatthewDaggitt:
Finally got it working. Will do some benchmarking for the paper now, but experimentally gives us close to the O(n) speedup predicted.


MatthewDaggitt:
Eek, this is a horrible bug. You get a different test behaviour on 9.4 vs 9.2 :scream:

MatthewDaggitt:
Okay so the order in which the elements of `HashTable.toList` are returned differs in 9.4 and 9.2 so we **need** to sort.

PR #606, opened 2023-06-28
First attempt at JSON backend

MatthewDaggitt:
Here's a first attempt at a JSON backend exposed under the new `vehicle compile --target JSON` option.
- All instance arguments resolved
- All implicit arguments erased

 In the end I did transform to a new datatype first, as I still refuse to believe that exporting our core datatype is the right approach, as it would mean it would become part of the external API.

At the moment, we're just fully re-exporting the standard libray alongside.


wenkokke:
@MatthewDaggitt Would you like to merge this, or would you like me to simply branch from this branch?

MatthewDaggitt:
Merge it 👍

PR #610, opened 2023-06-28
Fixes to testing framework

wenkokke:


wenkokke:
- [X] Extract testing framework out to standalone package
- [X] Change SomeOption to have one LocalOption constructor and one AdjustOption constructor, which call the respective functions.
- [X] Change --external-only to use safeReadBool and flagCLParser
- [ ] Change --external to take comma separated list
- [ ] Change --ignore-files and --ignore-lines to take comma separated lists
- [ ] Change --ignore-lines and --ignore-files to --ignore-matching-lines and --ignore-matching-files
- [ ] Change default comparison to hash digest.
- [ ] Change line length check into file size check. Suppress diff.
- [ ] Detect binary files using detectEncoding. Suppress diff.
- [ ] Add binary option akin to --ignore-matching-files. Suppress diff.

PR #613, opened 2023-06-30
Added pass to make functions pure and use it in JSON backend

MatthewDaggitt:
As discussed. The only wrinkle is that inferred parameters, are very difficult to infer here. Therefore, they too need to explicitly passed (at least for the moment).


wenkokke:
Aren't inferred parameters irrelevant? Therefore, we should be able to erase them, right?

MatthewDaggitt:
Definitely not irrelevant in the usual sense unfortunately. For example in the following:
```
@parameter(infer=True)
n : Nat

@dataset
data : Vector Rat n

@property
prop : Bool
prop = foreach i . data ! i > 0
```
Here, in Vehicle the `foreach` is definitely dependent on `n` as the value over which the variable `i` should range. So we can't just eliminate it. 

What we need do is some sort of dependency inversion pass, where the parameter `n` becomes derived from the dataset `data` object, rather than being used by it.

PR #636, opened 2023-07-12
Add specific constructor for partial applications in JSON AST

### LINKED ISSUE #443: Sort out naming of the two types of cons symbols
At the moment:
- List type constructor `::`
- Vector type function `::v`

Maybe `cons` and `::`? 

Then add to CHANGELOG.

### LINKED ISSUE #621: Shadowing bug in `Explicit` backend
When compiling the following standard library function to the "Explicit" backend we get a shadowing bug where the type of `xs` refers to the `f` bound in the function rather than the  `f` in the type-signature 
```
forallIn : forallT {f : Type -> Type} . {{HasFold f}} -> {{HasMap f}} -> (A -> Bool) -> f A -> Bool
forallIn f xs = bigAnd (map f xs)
```
Suspect it might be to do with the scoping hack in `Vehicle.Compile.FunctionaliseResources`.

(Had to disable something coercion elimination to work around this, so re-enable when fixed).

### LINKED ISSUE #628: Remove unnecessary currying from Python backend
Separate application nodes for partial applications of top-level declarations and builtins:
```hs
App        :: Expr -> [Expr] -> Expr
AppPartial :: Expr -> [Expr] -> Expr
```
For starters, we can say that AppPartial is only ever used with builtins, which would already clean up the generated Python tremendously.

However, the best case scenario is where we change Lam to match App in the exported AST:
```hs
Lam        :: [Binder] -> Expr -> Expr
```
That would let us bunch together abstractions, which lets us extend the use of AppPartial to applications of top-level declarations (i.e., uses of FreeVar), local declarations (i.e., uses of BoundVar) and anonymous functions (i.e, uses of Lam).

That would let us get rid of all pointless currying. It would let us automatically perform currying whenever an AppPartial is encountered, using the Python partial builtin. Furthermore, it would resolve the user-experience issue, where the user has to figure out the order of arguments for the generated top-level declarations. In the resulting function, the user will just be able to pass the arguments by name, as we indented. (edited)

MatthewDaggitt:
Gah, this was way more unpleasant than I'd hoped. However, it is now mostly working. The JSON AST has been moved to [vehicle/src/Vehicle/Expr/Relevant.hs](https://github.com/vehicle-lang/vehicle/compare/partial-apps?expand=1#diff-0ba9497f2a45ee10542e7ade48af1d70293702c6702ee6584769efad6b95c503) and you can find the new nodes there.

This isn't a finished PR. The `mnist-robustness` spec is unfortunately broken, and I haven't accepted accepted the golden tests. However it should give you something to work against @wenkokke. I'll continue to push fixes.


MatthewDaggitt:
This also fixes the problem that weren't producing valid python identifiers during monomorphisation.

MatthewDaggitt:
The outstanding bugs occur, because we don't normalise before checking for equality of implicits in monomorphisation.

wenkokke:
Does this mean types are normalised?

MatthewDaggitt:
No, they're not. Types are all still there. What made you think they might be erased?

wenkokke:
**NOTE**: All these tests are disabled in `test_python_load.py` and should be re-enabled before merging.

- [x] The files `vendor/vehicle/tests/golden/compile/simple-quantifierIn/spec.vcl` and `vendor/vehicle/tests/golden/compile/simple-let/spec.vcl` give:

    ```
    vehicle_lang._error.VehicleError:
      Internal scoping error during normalisation by evaluation:
        declaration 'StdLib.mapList' not found in scope...
    ```

- [x] The file `vendor/vehicle/tests/golden/compile/windController/spec.vcl` gives:

    ```
    vehicle_lang._error.VehicleError:
      Something went wrong internally.
      Please report the error shown below to `https://github.com/vehicle-lang/vehicle/issues`.
      Error: `FromNat` should have been removed after type-checking.
    ```

- [x] The file `vendor/vehicle/tests/golden/compile/mnist-robustness/spec.vcl` gives:

    ```
    vehicle_lang._error.VehicleError:
      Internal scoping error during normalisation by evaluation:
      the bound context of length '0' is smaller than the found DB index 𝓲3
    ```

wenkokke:
**NOTE**: Fixed. I've changed the Python code to use `BuiltinFunction` instead.

The new AST once again emits `Builtin` instead of `BuiltinOp`, which causes:
```
vehicle_lang.ast._decode.DecodeError:
  Could not decode value of type dict as Main:
    could not find class Builtin.
    (Main.declarations > DefFunction.type > Pi.binder > Binder.type > Pi.binder > Binder.type > App.func)
    {'tag': 'Builtin', 'contents': [{'tag': 'Provenance', 'contents': [2, 5, 2, 11]}, {'tag': 'VectorTyp
```

wenkokke:
The file `vendor/vehicle/tests/golden/compile/windController/spec.vcl` gives:
```
vehicle_lang._error.VehicleError:
  Something went wrong internally.
  Please report the error shown below to `https://github.com/vehicle-lang/vehicle/issues`.
  Error: `FromNat` should have been removed after type-checking.
```

wenkokke:
The file `vendor/vehicle/tests/golden/compile/mnist-robustness/spec.vcl` gives:
```
vehicle_lang._error.VehicleError:
  Internal scoping error during normalisation by evaluation:
  the bound context of length '0' is smaller than the found DB index 𝓲3
```

wenkokke:
Vehicle incorrectly emits a `PartialApp` for a fully applied if-statement, which results in the following errors:
```
FAILED tests/test_lossdl2_exec.py::test_loss_function_exec[test_indicator.vcl-samplers4-validate_output4] -
  AssertionError: assert 1.0 ==
    functools.partial(<bound method
      ABCBoolAsBoolBuiltins.If of PythonBuiltins(samplers={})>, False, 0.0, 1.0)
FAILED tests/test_lossdl2_exec.py::test_loss_function_exec[test_maximum.vcl-samplers5-validate_output5] - 
  AssertionError: assert 1.0 ==
    functools.partial(<bound method
      ABCBoolAsBoolBuiltins.If of PythonBuiltins(samplers={})>, False, 0.0, 1.0)
FAILED tests/test_lossdl2_exec.py::test_loss_function_exec[test_minimum.vcl-samplers6-validate_output6] -
  AssertionError: assert 0.0 ==
    functools.partial(<bound method
      ABCBoolAsBoolBuiltins.If of PythonBuiltins(samplers={})>, True, 0.0, 1.0)
```
Furthermore, Vehicle emits if-statements in the DL2 backend, which... seems like it shouldn't?

wenkokke:
There is a bug in the algorithm that Vehicle uses to detect partial applications. For instance, the application of `fold` in the following program is considered _total_, even though it isn't explicitly applied to any sequence:
```hs
bigAnd : forallT {f : Type -> Type} . {{HasFold f}} -> f Bool -> Bool
bigAnd = fold (\x y -> x and y) True
```
The ideal solution to this issue would be eta-expansion, which would let us translate `bigAnd` as:
```py
def bigAnd(sequence: Sequence[bool]) -> bool:
    return foldList(and_, True, sequence)
```
Another solution would be to correctly mark the application as _partial_, which would force us to translate `bigAnd` as:
```py
bigAnd = partial(foldList, and_, True)
```

MatthewDaggitt:
Okay, I've fixed several of the errors. The outstanding ones are still:
 - `vendor/vehicle/tests/golden/compile/mnist-robustness/spec.vcl` which we have some serious conceptual issues with inferable parameters.
 - emission of `if` statements. We need to discuss this, as I don't see anyway around emitting them at the moment.
 - the issue with `fold` which I really don't understand.

Willing to have a chat tonight?

MatthewDaggitt:
Also `pre-commit` is giving some typing issues.

MatthewDaggitt:
When running `pytest tests/test_python_exec.py` I now get _all_ the tests failing....
```
=========================================================== short test summary info ============================================================
FAILED tests/test_python_exec.py::test_loss_function_exec[test_addition.vcl-samplers0-validate_output0] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_at.vcl-samplers1-validate_output1] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_constant.vcl-samplers2-validate_output2] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_division.vcl-samplers3-validate_output3] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_indicator.vcl-samplers4-validate_output4] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_maximum.vcl-samplers5-validate_output5] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_minimum.vcl-samplers6-validate_output6] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_multiplication.vcl-samplers7-validate_output7] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_negation.vcl-samplers8-validate_output8] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_network.vcl-samplers9-network_validate_output] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_subtraction.vcl-samplers10-validate_output10] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_tensor.vcl-samplers11-validate_output11] - AttributeError: 'alias' object has no attribute 'asname'
FAILED tests/test_python_exec.py::test_loss_function_exec[test_variable.vcl-samplers12-validate_output12] - AttributeError: 'alias' object has no attribute 'asname'
============================================================== 13 failed in 1.11s ==============================================================
```

MatthewDaggitt:
Ah that was the print statement causing that....

MatthewDaggitt:
Okay I've fixed a small problem with the compilation of indices.

MatthewDaggitt:
Okay I've fixed the arity problems, but that's revealed a new problem. We need to eta-expand.
```
mapList : (Rat -> Rat) -> List Rat -> List Rat;
mapList f = fold (\ x -> \ xs -> f x :: xs) nil
```
is now correctly marked as having arity 2, but the lambda that's generated in the body only has one binder. Therefore when we try and pass two positional arguments to it in Python it falls over.

I'm afraid I'm getting on the plane now so won't be able to work for a while.

MatthewDaggitt:
We're there :tada:

PR #650, opened 2023-07-31
Refactored public facing Python bindings and added docstrings

MatthewDaggitt:
This PR is a proposal for what I think the user facing Python bindings should look like.

In particular the module `vehicle_lang` now exports `verify`, `generate_loss_function` and `generate_python_function`. These methods all add minimal assumptions about the format of the networks/datasets. For example:
 - `verify` expects filepaths to `.onnx` and `.idx` files
 - `generate_loss_function` and `generate_python_function` only assume network objects are callable and datasets are indexable.

I then envisage adding submodules `vehicle_lang.tensorflow`/`vehicle_lang.pytorch` that export specialised versions of these functions where,  all three functions expect actual tensorflow network objects and numpy datasets.

Thoughts @wenkokke ?


wenkokke:
I think that the names of the Python functions we create should mirror the Vehicle CLI.

So...
```python
vehicle_lang.compile(specification="path/to/spec.vcl", target=...)
```
...rather than...
```python
vehicle_lang.generate_loss_function(specification="path/to/spec.vcl", ...)
```

I think that the arguments to these function should mirror the flags for the Vehicle CLI.

MatthewDaggitt:
So that was my initial design, but I think there's honestly a fairly large benefit to abstracting away some of the complexity of the CLI. I'm not against having bindings to the general `compile` mode, but 99% of user's won't need to access it so I think presenting simplified interfaces like these are a good idea.

wenkokke:
> So that was my initial design, but I think there's honestly a fairly large benefit to abstracting away some of the complexity of the CLI. I'm not against having bindings to the general `compile` mode, but 99% of user's won't need to access it so I think presenting simplified interfaces like these are a good idea.

I don't want the bindings to the `compile` command to have the exact same result as calling `compile` from the CLI—i.e., produce JSON—but I want them to have the same command and argument names.

MatthewDaggitt:
> I don't want the bindings to the compile command to have the exact same result as calling compile from the CLI—i.e., produce JSON—but I want them to have the same command and argument names.

I agree, _when_ we add the general purpose bindings to the compile command, it should be named `compile` and should have the same argument names.

What I'm saying is that that is not what _these_ bindings are. This is the simplified interface that 99% of users are going to use.

So I think we're in agreement?

MatthewDaggitt:
Okay, I've added a `compile` function as requested, whose arguments mirror the CLI arguments.

wenkokke:
> > I don't want the bindings to the compile command to have the exact same result as calling compile from the CLI—i.e., produce JSON—but I want them to have the same command and argument names.
> 
> I agree, _when_ we add the general purpose bindings to the compile command, it should be named `compile` and should have the same argument names.
> 
> What I'm saying is that that is not what _these_ bindings are. This is the simplified interface that 99% of users are going to use.
> 
> So I think we're in agreement?

No, I believe that the `compile` function _should_ mirror the CLI arguments for `compile` _and_ should by default return a compiled Python module, rather than Vehicle or JSON output.

MatthewDaggitt:
> No, I believe that the compile function should mirror the CLI arguments for compile and should by default return a compiled Python module, rather than Vehicle or JSON output.

But it can't do that? How would it do it for the `Marabou` target? Or the `Agda` target?

MatthewDaggitt:
Do you want me to halt the merge?

wenkokke:
> > No, I believe that the compile function should mirror the CLI arguments for compile and should by default return a compiled Python module, rather than Vehicle or JSON output.
> 
> But it can't do that? How would it do it for the `Marabou` target? Or the `Agda` target?

It wouldn't? Only the loss function targets would allow the optional, on-by-default argument that says they should return Python modules.

MatthewDaggitt:
> It wouldn't? Only the loss function targets would allow the optional, on-by-default argument that says they should return Python modules.

At that point, isn't it _far clearer_ to have special purpose functions along the lines of what we have now? Rather than having mutable return types?

wenkokke:
> > It wouldn't? Only the loss function targets would allow the optional, on-by-default argument that says they should return Python modules.
> 
> At that point, isn't it _far clearer_ to have special purpose functions along the lines of what we have now? Rather than having mutable return types?

Fair. Why do you want these functions to be called "generate", though?

MatthewDaggitt:
> Fair. Why do you want these functions to be called "generate", though?

No strong reason. I'm open to alternative suggestions? Would `compile_to_loss_function` be better?

wenkokke:
> > Fair. Why do you want these functions to be called "generate", though?
> 
> No strong reason. I'm open to alternative suggestions? Would `compile_to_loss_function` be better?

Much happier. I'm doing some basic refactoring now, I'll add that.

MatthewDaggitt:
> Much happier. I'm doing some basic refactoring now, I'll add that.

Thanks!

wenkokke:
Closing in favour of #657

PR #658, opened 2023-08-01
Setup basic support for autodoc

wenkokke:


wenkokke:
Forgot to push before merging, so more commits need to be merged.

PR #681, opened 2023-08-16
Implement PGD for tensorflow backend

MatthewDaggitt:
The crucial change is that we need some notion of the domain of the variable from which to sample and clip into. This really should be derivable from the specification, but it's not easy to do (will discuss in meeting). At the moment therefore the user is expected to pass the domain explicitly. Very easy to do though so not a huge imposition.

The major downside is that currently this is _incredibly_ slow. Takes ~4 minutes to evaluate the MNIST robustness specification. Need to investigate why...


vehicle-python/src/vehicle_lang/compile/abc.py
```diff
@@ -32,10 +44,19 @@ class Builtins(
         _Nat,
         _Int,
         _Rat,
+        _Variable,
+        _VariableValue,
```
wenkokke: We really shouldn't be adding `_VariableValue`. Variable values will always be tensors of `_Nat`, `_Int`, or `_Rat`, and we simply don't have that typing information on the Python side of things, so we should simply treat them as `Vector[Any]`, instead of adding these arguments.

vehicle-python/src/vehicle_lang/compile/python.py
```diff
@@ -59,14 +67,63 @@
     "load_loss_function",
 ]
 
+
+################################################################################
+## Variable domain
+################################################################################
+
+
+class VariableDomain(AbstractVariableDomain[np.ndarray]):
```
wenkokke: We really shouldn't be mentioning numpy in the standard Python compilation. This should once more just be `Vector`.

vehicle-python/src/vehicle_lang/compile/abc.py
```diff
@@ -32,10 +44,19 @@ class Builtins(
         _Nat,
         _Int,
         _Rat,
+        _Variable,
```
wenkokke: Why would we abstract over `_Variable`? Presumably, this will always be `str`?

wenkokke (dismissed):

MatthewDaggitt:
After the latest merge, the Pygments tests are failing. Any chance you can take a look at them @wenkokke? Once we've fixed those, we should merge this in and open a new one for future improvements.

MatthewDaggitt:
Closing this PR as it's becoming hopelessly stale.

PR #780, opened 2024-02-10
Use new runner with macOS Sonoma

wenkokke:


wenkokke:
Blocked on the following change in Python 3.12:
IMG_3335

PR #790, opened 2024-03-26
Try to fix CI failure

MatthewDaggitt:


wenkokke:
The fix is probably to just set an upper bound for TensorFlow instead, at least for now.

MatthewDaggitt:
Hopefully got the fix sorted now. But if this fails, then yes I'll restrict the tensorflow version.

Issue #819, opened 2024-06-14
CITATION.cff validator is failing but not producing any useful output.


MatthewDaggitt:
@wenkokke any idea where the problems identified by the validator can be found?

wenkokke:
There's a bug in the BumpVer configuration that bumps Vehicle's version as well as the CFF version.

PR #820, opened 2024-06-14
Update CITATION.cff for latest release

MatthewDaggitt:


MatthewDaggitt:
Would be good if we could fix #819 as well in this one.

CONTRIBUTING.md
```diff
@@ -122,6 +122,11 @@ We recommend you install GHC and Cabal using [GHCup].
 
 If you'd like to use a different version of GHC, you can find the list of versions that we test with in [build-vehicle.yml]. However, be aware that building the Python bindings _requires_ [our preferred version](#the-preferred-version-of-ghc).
 
+Some of the Haskell libraries used by Vehicle also required `gmp` arithmetic library to be installed on the system. If it is not installed on your system you can install it, e.g. on Linux by
+   ```sh
+   sudo apt-get install libgmp3-dev
```
wenkokke, line 127: When are folks encountering this issue? It shouldn't be with a Python install.

wenkokke: This install command is specific to Ubuntu, which means it's not all that useful. Furthermore, most Haskell installation instructions inside instructions for installing gmp, or do it automatically.

CITATION.cff
```diff
@@ -79,6 +79,18 @@ references:
     year: "2022"
     url: "https://arxiv.org/abs/2202.05207"
 identifiers:
+  - type: url
+    value: "https://github.com/vehicle-lang/vehicle/releases/tag/v0.14.1"
+    description: "The GitHub release URL of tag v0.14.1."
+  - type: url
+    value: "https://gpypi.org/project/vehicle-lang/0.14.1/"
```
wenkokke, line 86: This shouldn't say "gpypi".

CITATION.cff
```diff
@@ -79,6 +79,18 @@ references:
...
+    description: "The PyPI release URL of version 0.14.1."
+  - type: url
+    value: "https://github.com/vehicle-lang/vehicle/releases/tag/v0.14.0"
+    description: "The GitHub release URL of tag v0.14.0."
+  - type: url
+    value: "https://gpypi.org/project/vehicle-lang/0.14.0/"
```
wenkokke, line 92: This shouldn't say "gpypi".

wenkokke (changes requested):

CONTRIBUTING.md
MatthewDaggitt: > When are folks encountering this issue? It shouldn't be with a Python install.

This was installing it from source on my new work laptop...

PR #850, opened 2024-10-11
Altered loss function compiler end to output an actual sane tensor-based language

MatthewDaggitt:
Massive refactoring, but the upshot is that we actually output an AST to Python that is much more trivial to convert. No more generic folds or maps, no more partially applied `App` nodes etc.

Also going to make domain analysis much easier.


wenkokke:
@MatthewDaggitt 

```
src/Vehicle/Backend/LossFunction/TensorCompilation.hs:11:48: error:
Error:     Module ‘Control.Monad.Reader’ does not export ‘void’
   |
11 | import Control.Monad.Reader (MonadReader (..), void)
   |                                                ^^^^
```

PR #856, opened 2024-10-15
Remove closure type parameter from Value

MatthewDaggitt:
Used to think we needed a full normal form, not just a weak normal form, for loss function translation, but as of a few months ago we no longer need it.


MatthewDaggitt:
I don't understand why this passed the merge queue when the tests failed?!

PR #870, opened 2024-12-12
Added timeout handling and stopped individual property errors being fatal

### LINKED ISSUE #869: Support for timeout
Vehicle doesn't currently seem to support specifying a timeout for verification (or at least I didn't find it in the docs), so I've tried to use `--verifier-args "--timeout=120"`.

I'm not sure what Marabou returns exactly (unknown? timeout?) but it is something Vehicle doesn't expect:

```
Verifying properties:
  robust!0 [.......................................................] 0/1 queries

Error: Unexpected output from the Marabou verifier. Cannot find 'sat' or 'unsat'
```

I could work around that, but unfortunately, when trying to verify a property on more than one image, verification will not continue afterwards. 

Expected behaviour would be an option for / handling of something like "verify this property on x images, with a timeout of y seconds for each".

MatthewDaggitt:
Fixes https://github.com/vehicle-lang/vehicle/issues/869 for @tflinkow. Errors are no longer fatal and adds some timeout detection. Can't actually test the timeout detection as I can't get Marabou to actually timeout (see comment in issue), but the worst that will happen is that Vehicle will treat it as an error which isn't disastrous.


MatthewDaggitt:
Also added printing of verification stats as per request.

Issue #879, opened 2025-03-25
Suggestion: python bindings to return str rather than print to stdout


Allenator1:
Hello,

We are currently in the midst of developing a GUI for vehicle. In this context, having the Python functions return a string to the caller would be more convenient than having to redirect from stdout to get that information. 

Thank you

wenkokke:
Could you elaborate on what functions you are referencing?

Allenator1:
The main function is vehicle_lang.verify

We would also like to eventually run `vehicle compile --target MarabouQueries` to show the compilation output, and `vehicle check` for error higlighting.

MatthewDaggitt:
So as Vehicle is a Haskell executable, all output (whether errors or standard output) has to be performed either via files or over `stdout`/`stderr`.  

Currently, the Python `Session` class already handles the capturing of `stdout`/`stderr` for us, e.g. 
https://github.com/vehicle-lang/vehicle/blob/f8dcd995b07d5c6178d10829c0d533cea7def2f0/vehicle-python/src/vehicle_lang/verify/__init__.py#L55

Therefore we already have the infrastructure necessary to access strings representing Vehicle's output. Does that make sense or I am misunderstanding the question?

Allenator1:
Hi Matthew,

Currently, it looks like the Session.check_output function doesn't correctly redirect from stdout. I think this is because it uses contextlib.redirect_stdout which redirects from the Python stdout rather than the C stdout. A possible solution might be to redirect C stdout (sys.stdout) using os.dup2

MatthewDaggitt:
Thanks for the explanation @Allenator1! 

@wenkokke does this match your understanding of the code you wrote all those years ago? @Allenator1 does using `os.dup2` work?

Allenator1:
@MatthewDaggitt 

Yes, os.dup2 works! However, I had to set up a pseudo terminal using pty.openpty() as vehicle seems to have ambigious (partial or no) output when redirecting to a pipe (I can set up a seperate issue for this).

Issue #877, opened 2025-03-19
Extend stable Python API


developing-ar:
The current Python API is declared [here](https://vehicle-lang.readthedocs.io/en/stable/python-api.html).
Use python bindings to expose the functionality of vehicle according to the interface defined in the [Unstable Python Api](https://vehicle-lang.readthedocs.io/en/stable/dev/unstable-python-api.html).

wenkokke:
Could you elaborate on why?

MatthewDaggitt:
We'd like to be able access the full suite of commands from Python, e.g. `verify`, `typeCheck` etc.

Does that make sense?

wenkokke:
Sure does. You already can, though, by simply calling `main`?

The following two unstable definitions can become stable:

- vehicle_lang.pygments.VehicleLexer
- vehicle_lang.verify.verify

There currently isn't any binding for type checking.

developing-ar:
We can try implement that, and `export` as well.
It would be helpful to get the syntax errors and counter examples in JSON with python bindings (but I'll create a separate issue for that).

MatthewDaggitt:
Yes, the idea is that we wrap calls to `main` in python functions that take the right arguments etc.

MatthewDaggitt:
This was achieved in v0.17

PR #883, opened 2025-04-06
Extend stable Python API

developing-ar:
# Description of Changes
- #877: python bindings for `compile`, `export`, `validate` and `check` using implemented infrastructure and only adding `__init__.py` for each function
- #879: uses `os.dup2()` to redirect correctly (for python bindings)

### DIFF
--- a/vehicle-python/src/vehicle_lang/validate/__init__py
+++ b/vehicle-python/src/vehicle_lang/validate/__init__py
@@ -1,0 +1,26 @@
+from pathlib import Path
+from typing import Union
+
+from .. import session
+from ..error import VehicleError
+
+def validate(
+    cache: Union[str, Path]
+) -> str:
+    """
+    Validate a verification result to check whether it still holds.
+
+    :param cache: The path to the proof cache used by Vehicle.
+    """
+    args = ["validate", "--cache", str(cache)]
+
+    # Call Vehicle
+    exc, out, err, _ = session.check_output(args)
+
+    # Check for errors
+    if exc != 0:
+        raise VehicleError(f"{err}")
+    elif not out:
+        return ''
+
+    return out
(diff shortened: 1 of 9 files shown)


vehicle-python/src/vehicle_lang/validate/__init__py
```diff
@@ -0,0 +1,26 @@
+from pathlib import Path
+from typing import Union
```
MatthewDaggitt: The file name for this isn't quite right.

MatthewDaggitt (changes requested):
Otherwise looks really good, thank you!

vehicle-python/src/vehicle_lang/validate/__init__py
developing-ar: Fixed in [36f728f](https://github.com/vehicle-lang/vehicle/pull/883/commits/36f728f40e2de99fc90366d41587434ef92e4f2a).

MatthewDaggitt (approved):

wenkokke:
@developing-ar Please remove the slave terminology. Otherwise this looks good.

@MatthewDaggitt It dawns on me that this is a rewrite I proposed some time ago, and which you shot down due to worries of maintaining the CLI interface twice. Is the interface stable enough now that you're no longer worried by this?

MatthewDaggitt:
@wenkokke sorry, it was rude of me to merge it so quickly! I won't merge any more PRs from @developing-ar without your approval as well. 

Yes, it was! However, yes the interface is pretty stable now. I haven't seriously changed it in about 2 years so I think it's reasonable to do. @developing-ar and others are also building on top of it on now, so it's about to become more stable out of necessity, even if it wasn't before!

developing-ar:
Sorry about that. Will make sure to double check terminology for upcoming PRs. I've opened a separate PR to fix that, or would you prefer reverting the commit and I open another PR accordingly?

wenkokke:
> Sorry about that. Will make sure to double check terminology for upcoming PRs. I've opened a separate PR to fix that, or would you prefer reverting the commit and I open another PR accordingly?

Fixing it in another PR is fine, thanks!

PR #885, opened 2025-04-08
List the resources (by declaration) in a vehicle specification

developing-ar:
# Description of changes
Added the list (top-level) option and sub-command for resources (networks, datasets, parameters), and properties in both a human readable and json format.

Usage:
```bash
$ vehicle list <resources|properties> -s <spec-file> <-j>
```

This is so that these elements can be rendered as "inputs" on the GUI.

### DIFF
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -42,9 +42,10 @@ import Options.Applicative
     switch,
     value,
   )
-import Vehicle.Backend.Prelude (DifferentiableLogicID, ITP, Target (..), TypingSystem (..), findTarget)
+import Vehicle.Backend.Prelude (DifferentiableLogicID, ITP, ListableEntities (..), Target (..), TypingSystem (..), findTarget)
 import Vehicle.Compile (CompileOptions (..))
 import Vehicle.Export (ExportOptions (..))
+import Vehicle.List (ListOptions (..))
 import Vehicle.Prelude
   ( Doc,
     enumerate,
@@ -94,6 +95,7 @@ data ModeOptions
   | Verify VerifyOptions
   | Validate ValidateOptions
   | Export ExportOptions
+  | List ListOptions
   deriving (Eq, Show)
 
 --------------------------------------------------------------------------------
@@ -167,6 +169,7 @@ modeOptionsParser =
         <> command "verify" verifyParserInfo
         <> command "validate" validateParserInfo
         <> command "export" exportParserInfo
+        <> command "list" listParserInfo
 
 --------------------------------------------------------------------------------
 -- Check mode
@@ -187,6 +190,27 @@ typeCheckParser =
 
 typeCheckParserInfo :: ParserInfo ModeOptions
 typeCheckParserInfo = info (Check <$> typeCheckParser) typeCheckDescription
+
+--------------------------------------------------------------------------------
+-- List mode
+
+listDescription :: InfoMod ModeOptions
+listDescription =
+  progDesc $
+    "List entities for a "
+      <> specificationFileExtension
+      <> " specification file"
+      <> "."
+
+listParser :: Parser ListOptions
+listParser =
+  ListOptions
+    <$> specificationParser
+    <*> listEntitiesParser
+    <*> outputAsJSONParser
+
+listParserInfo :: ParserInfo ModeOptions
+listParserInfo = info (List <$> listParser) listDescription
 
 --------------------------------------------------------------------------------
 -- Compile mode
@@ -295,6 +319,11 @@ allVerifiersFormats = map show (enumerate @QueryFormatID)
 allLossFunctionDLs :: [String]
 allLossFunctionDLs = map show (enumerate @DifferentiableLogicID)
 
+allListableEntities :: [Doc a]
+allListableEntities = flip map (enumerate @ListableEntities) $ \case
+  ExternalResources -> "i) Resources - (default) list all the networks, datasets, and parameters in the specification"
+  Properties -> "ii) Properties - list all the properties in the specification"
+
 allTargets :: [String]
 allTargets = allLossFunctionDLs <> allVerifiersFormats <> allITPs
 
@@ -375,6 +404,24 @@ typeSystemParser =
               )
         )
       <> value Standard
+
+listEntitiesParser :: Parser ListableEntities
+listEntitiesParser =
+  option auto $
+    long "listEntities"
+      <> short 'l'
+      <> help
+        ( "Which Entities are listed. "
+            <> layoutAsString
+              ( line
+                  <> line
+                  <> indent
+                    2
+                    ( vsep allListableEntities
+                    )
+              )
+        )
+      <> value ExternalResources
 
 specificationParser :: Parser FilePath
 specificationParser =
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -205,8 +205,8 @@ listDescription =
 listParser :: Parser ListOptions
 listParser =
   ListOptions
-    <$> specificationParser
-    <*> listEntitiesParser
+    <$> listModeParser
+    <*> specificationParser
     <*> outputAsJSONParser
 
 listParserInfo :: ParserInfo ModeOptions
@@ -319,11 +319,6 @@ allVerifiersFormats = map show (enumerate @QueryFormatID)
 allLossFunctionDLs :: [String]
 allLossFunctionDLs = map show (enumerate @DifferentiableLogicID)
 
-allListableEntities :: [Doc a]
-allListableEntities = flip map (enumerate @ListableEntities) $ \case
-  ExternalResources -> "i) Resources - (default) list all the networks, datasets, and parameters in the specification"
-  Properties -> "ii) Properties - list all the properties in the specification"
-
 allTargets :: [String]
 allTargets = allLossFunctionDLs <> allVerifiersFormats <> allITPs
 
@@ -405,23 +400,21 @@ typeSystemParser =
         )
       <> value Standard
 
-listEntitiesParser :: Parser ListableEntities
-listEntitiesParser =
-  option auto $
-    long "listEntities"
-      <> short 'l'
-      <> help
-        ( "Which Entities are listed. "
-            <> layoutAsString
-              ( line
-                  <> line
-                  <> indent
-                    2
-                    ( vsep allListableEntities
-                    )
-              )
+listModeParser :: Parser ListableEntities
+listModeParser =
+  hsubparser $
+    command
+      "resources"
+      ( info
+          (pure ExternalResources)
+          (progDesc "List all networks, datasets, and parameters in the specification.")
+      )
+      <> command
+        "properties"
+        ( info
+            (pure Properties)
+            (progDesc "List all properties in the specification.")
         )
-      <> value ExternalResources
 
 specificationParser :: Parser FilePath
 specificationParser =
--- a/vehicle/src/Vehicle/List.hs
+++ b/vehicle/src/Vehicle/List.hs
@@ -14,8 +14,8 @@ import Vehicle.Prelude.Logging.Instance
 import Vehicle.TypeCheck (TypeCheckOptions (..), runCompileMonad, typeCheckUserProg)
 
 data ListOptions = ListOptions
-  { specification :: FilePath,
-    listEntities :: ListableEntities,
+  { listEntities :: ListableEntities,
+    specification :: FilePath,
     outputAsJSON :: Bool
   }
   deriving (Eq, Show)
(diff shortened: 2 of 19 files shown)


vehicle/src/Vehicle/CommandLine.hs
```diff
@@ -454,6 +455,13 @@ outputAsJSONParser =
       <> help "Output the program as JSON instead of text."
       <> internal
 
+resourcesParser :: Parser Bool
+resourcesParser =
+  switch $
+    long "resources"
+      <> short 'r'
+      <> help "Output the resources outlined in a vehicle file. Only prints if typeCheck is successful."
```
MatthewDaggitt: Maybe "After type-checking output the resources (e.g. networks, parameters, datasets) in the file as JSON"?

vehicle/src/Vehicle/TypeCheck.hs
```diff
@@ -140,6 +143,22 @@ printPropertyTypes (Main decls) = do
       let propertyType = prettyFriendlyEmptyCtx (typeOf decl)
       propertyName <+> ":" <+> propertyType
 
+printResources :: (MonadIO m, MonadCompile m, PrintableBuiltin builtin) => Prog builtin -> m ()
+printResources (Main decls) = do
+  let networks = filterByAbstractSort NetworkDef decls
```
MatthewDaggitt: So this function definitely works, nice job! I have a couple of suggested improvements though.

1. We'd probably like the resources to be returned in the same order that they were in the file so that they can be displayed in that order in the GUI? 
2. For the GUI, we probably need the type information as well, not just the name, in order to create the right input boxes and do validation?
3. It would be good if we actually output proper JSON rather than hand-crafting a format?

Doing 2 is hard so maybe we'll leave that for another issue, but we should be able to do 1 and 3 pretty easily. Instead of filtering, can you write a function that recurses over all the declarations and returns something of type `[AbstractSort, Name]`. You then should be able to convert that list directly to JSON with `toJSON`?

Also FYI you need to capture `NonInferable` parameters as well.

MatthewDaggitt (changes requested):
Great job! I have a couple of suggestions for improvements, but it would definitely work!

MatthewDaggitt:
Also looks like the help text has changed (see failing test in CI), you will need to update the golden test file. Instructions here: https://github.com/vehicle-lang/vehicle/blob/dev/CONTRIBUTING.md#updating-the-golden-files

wenkokke:
If the semantics of this flag is that it prints the resources and then exits, please transform it into a command named "list-resources".

vehicle/src/Vehicle/TypeCheck.hs
wenkokke: @MatthewDaggitt did you intend for that to be a list of tuples?

wenkokke: I agree, this looks great! I have a few more suggestions:

The filtering, as well as any other nontrivial operation on the data, should be done externally, rather than in Vehicle, otherwise we'll have to keep modifying Vehicle as your needs change, or we'll end up with the logic split between two places.

The output should be JSON, as it's intended to be consumed programmatically. However, I'd be even happier if the `list-resources` command took an optional `--json` flag that changes its output form human readable to JSON in line with how the other commands work.

Contrary to Matthew's suggestion, I believe the names and sorts of the resources should be output as  a list of objects, so that future additions won't break external tools. This requires a little bit more work, defining a Haskell type and its ToJSON instance.

vehicle/src/Vehicle/CommandLine.hs
```diff
@@ -184,6 +184,7 @@ typeCheckParser =
   TypeCheckOptions
     <$> specificationParser
     <*> typeSystemParser
+    <*> resourcesParser
```
wenkokke: While it requires type checking, `list-resources` should be its own top level command which outputs the resources as well as other info. It should take a `--json` flag that changes its output from a human readable list to JSON.

wenkokke (changes requested):

vehicle/src/Vehicle/TypeCheck.hs
developing-ar: @wenkokke that makes sense for haskell type to have the its ToJSON instance. So, for the 
> names and sorts of the resources should be output as a list of objects

that means that the returned JSON is a "list" of all the resources in the file and each object has its name and "sort" (and later on type).

@MatthewDaggitt type information does not seem to difficult to extract as it is already done in `printProperties`, unless you were expecting a more sophisticated "typing".

wenkokke: I think @MatthewDaggitt was thinking about outputting the types as structured JSON. However, it's probably fine to add the printed type as a string, if you'd like to show this to the user. However, it may be a bit tricky to get the type in the right shape and get the right printing function.

MatthewDaggitt: Yes, the type is tricky as we can't convert raw vehicle types of JSON (and nor is it clear we want to). For now, yes we could just `prettyFriendly` the type and attach it to the tuple. @wenkokke is of course right that I meant `[(AbstractSort, Name)]` or now `[(AbstractSort, Name, Text)]` for the type.

MatthewDaggitt: I agree with @wenkokke as well, maybe splitting this out under a separate `resources` command mode is the way to go. (I prefer `resources` over `list-resources`) and having the `--json` flag as well.

wenkokke: I think `list-resources` might be better since I anticipate that it will expand to include more list commands, rather than expand to include more resource related commands.

wenkokke: > Yes, the type is tricky as we can't convert raw vehicle types of JSON (and nor is it clear we want to). For now, yes we could just `prettyFriendly` the type and attach it to the tuple. @wenkokke is of course right that I meant `[(AbstractSort, Name)]` or now `[(AbstractSort, Name, Text)]` for the type.

It should be a list of a custom type, rather than a tuple, so that we can have a custom ToJSON instance, as mentioned above.

MatthewDaggitt: Ooh like what?

MatthewDaggitt: But okay, yup `list-resources` is fine!

wenkokke: > Ooh like what?

```json
{
  "name": "...",
  "sort": "...",
  "type": "..."
}
```

MatthewDaggitt: Sorry, I meant "What other things will we want to list in the future?"

wenkokke: Properties. Functions. Implicit variables. Who knows?

vehicle/src/Vehicle/CommandLine.hs
developing-ar: Sounds good. Will do!

developing-ar:
@wenkokke @MatthewDaggitt then, how about a new top-level command `list` and then `--resources`, `--properties` etc. (and `--json`) can all be command-line options?

MatthewDaggitt:
> @wenkokke @MatthewDaggitt then, how about a new top-level command `list` and then `--resources`, `--properties` etc. (and `--json`) can all be command-line options?

Yup, sounds good to me!

wenkokke:
> @wenkokke @MatthewDaggitt then, how about a new top-level command `list` and then `--resources`, `--properties` etc. (and `--json`) can all be command-line options?

I'd prefer them to be subcommands, in that case, because they should be exclusive, so:

`vehicle list resources`

and:

`vehicle list resources --json`

developing-ar:
> > @wenkokke @MatthewDaggitt then, how about a new top-level command `list` and then `--resources`, `--properties` etc. (and `--json`) can all be command-line options?
> 
> I'd prefer them to be subcommands, in that case, because they should be exclusive, so:
> 
> `vehicle list resources`
> 
> and:
> 
> `vehicle list resources --json`

Done! Although its `Resources` instead of `resources`. I can update that if the case is important. I've added a `Properties` option as well.

wenkokke:
The case is important. All commands should be lowercase, for various reasons. Please do change it to lowercase.

developing-ar:
> The case is important. All commands should be lowercase, for various reasons. Please do change it to lowercase.

@wenkokke Sorry, I forgot to clarify that I'm using it as an option rather than a sub-command.
For example, in typeCheck:
```
-t <Standard|Linearity|Polarity>
```
in List:
```
-l <Resources|Properties>
```
I can convert it to a sub-command instead, and use the lower case accordingly - unless it's fine to have them as options.

wenkokke:
> > The case is important. All commands should be lowercase, for various reasons. Please do change it to lowercase.
> 
> 
> 
> @wenkokke Sorry, I forgot to clarify that I'm using it as an option rather than a sub-command.
> 
> For example, in typeCheck:
> 
> ```
> 
> -t <Standard|Linearity|Polarity>
> 
> ```
> 
> in List:
> 
> ```
> 
> -l <Resources|Properties>
> 
> ```
> 
> I can convert it to a sub-command instead, and use the lower case accordingly - unless it's fine to have them as options.

I'd prefer as lowercase commands. Firstly, because always lowercase. Secondly, because I'd ideally like options to have globally unique names (ie for every subcommand) to avoid confusion, and would prefer not to push us further from that ideal.

vehicle/src/Vehicle/List.hs
```diff
@@ -0,0 +1,67 @@
...
+  { entitySort :: Text,
+    entityName :: Text,
+    entityType :: Text
+  }
+  deriving (Eq, Show, Generic)
+
+instance ToJSON ListEntity
+
+instance Pretty ListEntity where
+  pretty listEntity = pretty (entitySort listEntity) <+> pretty (entityName listEntity) <+> pretty (entityType listEntity)
```
MatthewDaggitt: Can we avoid over long lines, maybe:
```haskell
pretty listEntity = 
  pretty (entitySort listEntity) <+> 
  pretty (entityName listEntity) <+> 
  pretty (entityType listEntity)
```

vehicle/src/Vehicle/List.hs
```diff
@@ -0,0 +1,67 @@
...
+  (imports, typedProg) <-
+    typeCheckUserProg $
+      TypeCheckOptions
+        { specification = specification,
+          typingSystem = Standard
+        }
+  let mergedProg = mergeImports imports typedProg
+  printResources mergedProg listEntities outputAsJSON
+
+printResources :: (MonadIO m, MonadCompile m, PrintableBuiltin builtin) => Prog builtin -> ListableEntities -> Bool -> m ()
```
MatthewDaggitt: Again can we lay this out as:
```haskell
printResources :: 
  (MonadIO m, MonadCompile m, PrintableBuiltin builtin) => 
  Prog builtin -> 
  ListableEntities -> 
  Bool -> 
  m ()
```

vehicle/src/Vehicle/List.hs
```diff
@@ -0,0 +1,67 @@
...
+printResources (Main decls) listEntities outputAsJSON = do
+  let filteredDecls =
+        if listEntities == ExternalResources
+          then filter isAbstractDecl decls
+          else filter isPropertyDecl decls
```
MatthewDaggitt: This could be written more succinctly as:
```haskell
let filterFn = case listEntities of
  ExternalResources -> isAbstractDecl
  Properties -> isPropertyDecl
let filteredDecls = filter filterFn decls
```
This has two advantages. Firstly reduces the duplication of `filter ... decls`. Secondly the compiler will warn if we add a new option to `ListableEntities`.

vehicle/src/Vehicle/List.hs
```diff
@@ -0,0 +1,67 @@
...
+  let listDecls = fmap (\decl -> convertDeclToListEntity decl (if listEntities == ExternalResources then pack $ show $ pretty (abstractSortOf decl) else "@property")) filteredDecls
```
MatthewDaggitt: I don't think you need the `pack (show)` here as you are pretty printing it on line 42.

vehicle/src/Vehicle/List.hs
```diff
@@ -0,0 +1,67 @@
...
+  let outputDocs =
+        if outputAsJSON
+          then pretty $ unpack $ encodePretty' prettyJSONConfig $ toJSON listDecls
+          else pretty listDecls
```
MatthewDaggitt, line 57: Can extract the `pretty` in both branches to line 46.

MatthewDaggitt (changes requested):
Looks good! Some minor Haskell style pointers!

MatthewDaggitt:
Yup, a subcommand would be great!

vehicle/src/Vehicle/List.hs
developing-ar: Great! will do

vehicle/src/Vehicle/List.hs
developing-ar: It needs to be in `Text` for `convertDeclToListEntity()` but I've removed that and made the `convertDeclToListEntity` use `Doc a` instead.

vehicle/src/Vehicle/List.hs
developing-ar: `pretty` is applied differently for each branch (produced  types), and extracting `pretty` gives the error for `listDecls`
```
 Couldn't match type ‘ListEntity’ with ‘Char’
```
Unless I'm missing something...

MatthewDaggitt: Ah right, fair enough!

MatthewDaggitt (commented):
Looks good to me! Can you add a couple of golden tests for example specifications that have resources (e.g. mnist-robustness, acasXu)

developing-ar:
@MatthewDaggitt  I'm adding golden tests for acasXu, mnist-robustness, and wind-controller for both `list resources` and `list properties`. Would you like me to update all the tests for **all** the golden tests (examples) ?

vehicle/tests/golden/compile/acasXu/ListProperties.out.golden
```diff
@@ -0,0 +1,10 @@
+[ @property property1 Bool
```
MatthewDaggitt: Hmm, I was kind of expecting that this was going to look more like JSON:
```
[ { entitySort = @property,
 , entityName = property1,
 , entityType = Bool
} ,
...
]
```
This current format is going to be kind of hard to parse on the other side!

MatthewDaggitt (commented):
Yes this small set of tests is fine. It's not terribly complicated code at the end of the day, and I think you've covered all the edge cases.

It would be good if we could tweak the actual format to be JSON, but after that happy to merge in!

vehicle/tests/golden/compile/acasXu/ListProperties.out.golden
developing-ar: The JSON option does give this output format which should be fine to parse:
```
[
  {
    "entitySort": "@property",
    "entityName": "property1",
    "entityType": "Bool",
  },
 ...
]
```
I did not include the json option in the tests, I'll update that - I assumed the json option might not be part of the golden tests since it's not the "base" output, since a previous [discussion](https://github.com/vehicle-lang/vehicle/pull/885#discussion_r2034454551) was that the default output is a human-readable list, and the optional json option.

Unless you meant for changing the default format?

MatthewDaggitt: No, that's perfect! As this API is meant to be consumed by another program rather than a human, I think it makes most sense to have the golden test use JSON!

Thanks for the change!

MatthewDaggitt (approved):
I'm happy with it! @wenkokke ?

wenkokke (approved):

wenkokke:
Looks good to me!

Issue #886, opened 2025-04-16
AST load function does not parse JSON


Allenator1:
The AST load function is not able to parse JSON files generated by `vehicle compile --json`. 

This is because `ast/ __init__.py` is out of date, and does not represent the current version of the generated JSON

A fix would be to adjust the Python classes in `ast/ __init__.py`

wenkokke:
I believe that @MatthewDaggitt was still actively working on the representation on the Haskell side of things, and we decided not to update the Python library while the AST representation was still undergoing massive changes.

@MatthewDaggitt: Has this stabilized?

MatthewDaggitt:
Yes, I believe it has stabilized sufficiently to make this worthwhile.

MatthewDaggitt:
This was fixed in https://github.com/vehicle-lang/vehicle/pull/1003

PR #888, opened 2025-04-16
Issue #886: update ast/__init__.py to correctly parse JSON AST 

Allenator1:
Added classes and updated the fields for existing classes to correctly represent the JSON AST

### DIFF
--- a/vehicle-python/src/vehicle_lang/ast/__init__.py
+++ b/vehicle-python/src/vehicle_lang/ast/__init__.py
@@ -59,244 +59,96 @@ class Tensor(Generic[DType]):
     value: Tuple[DType, ...]
 
 
-################################################################################
-# Builtin Constants
-################################################################################
-
-
-@dataclass(frozen=True, init=False)
-class BuiltinConstant(AST):
-    def __init__(self) -> None:
-        raise TypeError("Cannot instantiate abstract class BuiltinConstant")
+#################################################################################
+# Dimensions
+#################################################################################
 
 
 @dataclass(frozen=True)
-class Unit(BuiltinConstant):
-    pass
-
-
-@dataclass(frozen=True)
-class NilList(BuiltinConstant):
-    pass
-
-
-################################################################################
-# Builtin Functions
-################################################################################
-
-
-@dataclass(frozen=True, init=False)
-class BuiltinFunction(AST):
-    def __init__(self) -> None:
-        raise TypeError("Cannot instantiate abstract class BuiltinFunction")
-
-
-@dataclass(frozen=True)
-class ConsList(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class NotBoolTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class AndBoolTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class OrBoolTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class NegRatTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class AddRatTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class SubRatTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class MulRatTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class DivRatTensor(BuiltinFunction):
-    pass
-
-
-@dataclass(frozen=True)
-class EqRatTensor(BuiltinFunction):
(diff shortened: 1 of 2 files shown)


vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
 
 
 @dataclass(frozen=True, init=False)
-class BuiltinFunction(AST):
+class BuiltinLiteral(AST):
+    # Abstract base for literal values
+    value: Any
+
+    def __init__(self) -> None:
+        raise TypeError("Cannot instantiate abstract class")
```
MatthewDaggitt, line 99: In each of these errors, can we say what that class is?

vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
+
+################################################################################
+# Builtin Constants
+################################################################################
 
 
 @dataclass(frozen=True)
-class SubRatTensor(BuiltinFunction):
+class Unit(BuiltinConstant):
     pass
```
MatthewDaggitt, line 191: This is the AST we're trying to mimic: https://github.com/vehicle-lang/vehicle/blob/dev/vehicle/src/Vehicle/Backend/LossFunction/JSON.hs

I don't think we need a `Unit` class now?

vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
 
 
 @dataclass(frozen=True)
-class MulRatTensor(BuiltinFunction):
-    pass
+class ConstTensor(BuiltinConstant):
+    literal: BuiltinLiteral
```
MatthewDaggitt, line 196: Doesn't have to be a literal, can be any arbitrary `Expression`.

vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
+    dimension: DimensionCons
```
MatthewDaggitt, line 197: Again, doesn't have to be a `Cons` constructor, can be any arbitrary `Expression`.

vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
 
 
 @dataclass(frozen=True)
-class DivRatTensor(BuiltinFunction):
+class NilList(BuiltinConstant):
```
MatthewDaggitt, line 201: No longer exists in AST

vehicle-python/src/vehicle_lang/ast/__init__.py
```diff
@@ -59,373 +59,459 @@ class Tensor(Generic[DType]):
...
-    pass
+class NegRatTensor(BuiltinFunction):
+    body: Sequence[FunctionInput]
 
 
 @dataclass(frozen=True)
-class NeIndex(BuiltinFunction):
-    pass
+class AddRatTensor(BuiltinFunction):
+    body: Sequence[FunctionInput]
```
MatthewDaggitt, line 287: So in the AST at https://github.com/vehicle-lang/vehicle/blob/dev/vehicle/src/Vehicle/Backend/LossFunction/JSON.hs this has two arguments both of which `JExpr`, so I'd expect this to be something like:
```
class AddRatTensor(BuiltinFunction):
    arg1: Expression1
    arg2: Expression2
```

MatthewDaggitt (changes requested):
Looks like a solid start, thank you!

I have some comments that link to the AST that we're trying to load in. You'll need to ensure that the code reflects that accurately. I haven't highlighted everywhere where it diverges, but hopefully you should be able to through and see where changes need to be made?

vehicle-python/src/vehicle_lang/ast/__init__.py
Allenator1: I see.. 

I'll apply fixed arguments for all the other functions as well (except for StackTensor they all seem to have fixed arity)

vehicle-python/src/vehicle_lang/ast/__init__.py
Allenator1: Looking at JSON.hs, would it be better to make Expression the parent class for BuiltinFunction, BuiltinLiteral, and BuiltinConstant?

MatthewDaggitt: I don't think we even need those classes anymore, as they don't appear in the AST!

PR #889, opened 2025-04-17
Issue 886: Update AST according to Matthew's suggestions

Allenator1:
- Removed redundant data classes and abstract classes.
- Changed ordering to follow structure in JSON.hs
- Added classes according to JSON.hs.
- Updated dependent modules so that pre-commit tests are successful

### DIFF
--- a/vehicle-python/src/vehicle_lang/compile/abc/translation.py
+++ b/vehicle-python/src/vehicle_lang/compile/abc/translation.py
@@ -64,36 +64,30 @@ class ABCTranslation(
     def translate_expression(
         self, expression: vcl_ast.Expression
     ) -> vcl_var.Expression:
-        if isinstance(expression, vcl_ast.App):
-            return self.translate_App(expression)
         if isinstance(expression, vcl_ast.Var):
             return self.translate_Var(expression)
-        if isinstance(expression, vcl_ast.Builtin):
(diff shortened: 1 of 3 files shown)


vehicle-python/src/vehicle_lang/compile/abc/translation.py
```diff
@@ -64,36 +64,30 @@ def translate_DefPostulate(
...
         if isinstance(expression, vcl_ast.Var):
             return self.translate_Var(expression)
-        if isinstance(expression, vcl_ast.Builtin):
-            return self.translate_Builtin(expression)
         if isinstance(expression, vcl_ast.Lam):
             return self.translate_Lam(expression)
-        if isinstance(expression, vcl_ast.PartialApp):
-            return self.translate_PartialApp(expression)
         if isinstance(expression, vcl_ast.Pi):
             return self.translate_Pi(expression)
```
MatthewDaggitt, line 72: Unfortunately here we'll have to add all the extra subclasses of `Expression` we've added above.

MatthewDaggitt (approved):
Otherwise looks good, and definitely a step in the right direction! I think you can delete all the commented out code.

MatthewDaggitt:
I think this has unfortunately been superseded by work by @ggustavs. Apologies @Allenator1

PR #890, opened 2025-04-17
Generate VNNLibQueries for (upcoming) version 2.0

developing-ar:
# Description of Changes
Pass the `metaNetwork` object to `compileQuery` for the VNNLib(2.0) such that declarations in a query are defined as:
```
(declare-network <networkName>
    (declare-input <networkName>X Real <shape>)
    (declare-output <networkName>Y Real <shape>))
```
With network name being: `<metaNetworkName>@<networkApplicationIndex>`.
Including tests for: acasXu, mnist-robustness, windcontroller and monotonicity.

Also, fixed the parentheses on a query's constraints (`assert ...`).


vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
-compileVariableDecl :: (MonadLogger m) => QueryVariable -> m (Doc a)
-compileVariableDecl var = return $ parens ("declare-fun" <+> pretty var <+> "() Real")
-
 compileAssertion :: (MonadLogger m) => QueryAssertion QueryVariable -> m (Doc a)
 compileAssertion QueryAssertion {..} = do
   let compiledRel = compileRel rel
   let compiledLHS = foldl compileCoefVar "" (NonEmpty.tail lhs)
   let compiledRHS = prettyRationalAsFloat rhs
-  return $ parens "assert" <+> parens (compiledRel <+> parens compiledLHS <+> compiledRHS)
+  return $ parens ("assert" <+> parens (compiledRel <+> parens compiledLHS <+> compiledRHS))
```
MatthewDaggitt, line 101: Nice spot!

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+-- | Compiles the declarations for a network query
+compileNetworkEntry :: MetaNetworkEntry -> Doc a
+compileNetworkEntry metaNetworkEntry = do
+  let networkInputDocs = indentQuery $ uncurry compileNetworkInput $ networkTensor metaNetworkEntry Input
+  let networkOutputDocs = indentQuery $ uncurry compileNetworkOutput $ networkTensor metaNetworkEntry Output
+  parens ("declare-network" <+> pretty (metaNetworkEntryName metaNetworkEntry) <> line <> networkInputDocs <> line <> networkOutputDocs)
+
+-- | Compiles "all" (network) entries in the MetaNetwork
+compileNetworks :: (MonadLogger m) => MetaNetwork -> m (Doc a)
+compileNetworks metaNetwork = return $ concatWith (\a b -> a <> line <> b) (fmap compileNetworkEntry metaNetwork)
```
MatthewDaggitt: I think `vsep` is a shorthand for `concatWith (\a b -> a <> line <> b)` ?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+compileNetworkInput :: Name -> TensorShape -> Doc a
+compileNetworkInput name shape = parens ("declare-input" <+> pretty name <+> "Real" <+> pretty (PrettyTensorShape shape))
+
+-- | Compile a network output
+compileNetworkOutput :: Name -> TensorShape -> Doc a
+compileNetworkOutput name shape = parens ("declare-output" <+> pretty name <+> "Real" <+> pretty (PrettyTensorShape shape))
+
+-- | Indents a query
+indentQuery :: Doc a -> Doc a
+indentQuery = indent 2
```
MatthewDaggitt: I think you can just inline this as `indent 2`?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+
+-- | Fetches the name of the network from the meta network entry
+networkName :: MetaNetworkEntry -> Name
+networkName = metaNetworkEntryName
+
+-- | "Generates" the name and fetches the shape of the the input or output network tensor
+networkTensor :: MetaNetworkEntry -> InputOrOutput -> (Name, TensorShape)
+networkTensor metaNetworkEntry inputOrOutput = do
```
MatthewDaggitt: Might be nice to use the record opening syntax `MetaNetworkEntry {..}` here, and then all the fields automatically come into scope and you can use them directly (e.g. `metaNetworkEntryInfo` rather than `metaNetworkEntryInfo metaNetworkEntry`)? Might be usable in a few other places as well.

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+compileNetworkOutput :: Name -> TensorShape -> Doc a
+compileNetworkOutput name shape = parens ("declare-output" <+> pretty name <+> "Real" <+> pretty (PrettyTensorShape shape))
+
+-- | Indents a query
+indentQuery :: Doc a -> Doc a
+indentQuery = indent 2
+
+-- | Fetches the name of the network from the meta network entry
+networkName :: MetaNetworkEntry -> Name
+networkName = metaNetworkEntryName
```
MatthewDaggitt: Once you are using record opening syntax, this function hopefully won't be needed?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
 
 -- | Compiles an individual variable
 compileVNNLibVar :: CompileQueryVariable
-compileVNNLibVar inputOrOutput ioIndex = do
+compileVNNLibVar metaNetworkEntry inputOrOutput ioIndex = do
   let name = if inputOrOutput == Input then "X" else "Y"
-  layoutAsText $ name <> "_" <> pretty ioIndex
+  layoutAsText $ pretty (networkName metaNetworkEntry) <> name <> "_" <> pretty ioIndex
```
MatthewDaggitt: Looking at this a thought has just occurred to me about the main VNNLib spec. Do we allow `_` characters in the variable names in the new VNNLib spec? We need to decide what are allowable variable names!

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+  case inputOrOutput of
+    Input -> (networkName metaNetworkEntry <> "X", dimensions $ inputTensor networkTensors)
+    Output -> (networkName metaNetworkEntry <> "Y", dimensions $ outputTensor networkTensors)
+
+-- | Compiles the declarations for a network query
+compileNetworkEntry :: MetaNetworkEntry -> Doc a
+compileNetworkEntry metaNetworkEntry = do
+  let networkInputDocs = indentQuery $ uncurry compileNetworkInput $ networkTensor metaNetworkEntry Input
+  let networkOutputDocs = indentQuery $ uncurry compileNetworkOutput $ networkTensor metaNetworkEntry Output
+  parens ("declare-network" <+> pretty (metaNetworkEntryName metaNetworkEntry) <> line <> networkInputDocs <> line <> networkOutputDocs)
```
MatthewDaggitt: I suspect some indentation around `networkInputDocs` and `networkOutputDocs` is needed?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -34,27 +36,59 @@ outputFormat =
...
+-- | Compiles "all" (network) entries in the MetaNetwork
+compileNetworks :: (MonadLogger m) => MetaNetwork -> m (Doc a)
+compileNetworks metaNetwork = return $ concatWith (\a b -> a <> line <> b) (fmap compileNetworkEntry metaNetwork)
 
 -- | Compiles an expression representing a single VNNLib query.
 compileVNNLibQuery :: CompileQuery
-compileVNNLibQuery _address (QueryContents variables assertions) = do
-  variableDocs <- forM variables compileVariableDecl
+compileVNNLibQuery _address (QueryContents _variables assertions) metaNetwork = do
+  variableDocs <- compileNetworks metaNetwork
```
MatthewDaggitt: Could rename this to `networkDocs`?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
MatthewDaggitt: Also it looks like you compute the name of the network variable both here and on lines 64/65. Could we refactor the computation of the variable name into a single place?

MatthewDaggitt (changes requested):
Looks good! Some minor comments, but nothing major. It would be great if you could add a few golden tests as well? You can use the existing specifications and just add a VNNLib entry.

MatthewDaggitt:
Ah there's one more snag in this scheme. You take the network name as the basis for the variable name which is good! However, it's going to go wrong for the monotonicity test, where we have two applications of the _same_ network and therefore the same name!

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
developing-ar: You're right! I'll fix that.

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
developing-ar: Definitely helpful - will do.

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
developing-ar: The "problem" is that assert and declare statements allow/require different variable names:
For assertions, the `_` character **is** allowed (for individual elements in the tensor `TensorElement`).
For declarations, the `_` character **is not** allowed (`VariableName`).

I'm not sure how that can be changed, because "here" the assertion variables are generated, and for the declarations on  lines 64-65. Ideas?

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
developing-ar: Do you mean:
1. an extra "line" of separation between `networkInputDocs` and `networkOutputDocs`, or 
2. applying the [indent](https://github.com/vehicle-lang/vehicle/pull/890#discussion_r2051952851), or 
3. adding a line after `networkOutputDocs` for the closing bracket to be on the next line

Or something else (which I have missed)?

developing-ar:
@MatthewDaggitt I see! Then "indexing" (the list of) networks to derive the network name might solve this? eg. network `"acasXu"` would become `"acasXu0"` and the inputs and outputs for this network are updated accordingly. 
Although this could make differently named networks the same name (if an existing network name is terminated with a number).

Maybe this can only be "generated" in contexts where **all** the networks in context have the same name, or have some other constraint on allowed network names (in vehicle)?

MatthewDaggitt:
> Then "indexing" (the list of) networks to derive the network name might solve this?

Yes, that would fix it. I'm not sure I can think of any other sensible way to do it.... Yes, good point, we do variable names to end with numbers in Vehicle. Maybe we need use `-` as a special character to separate the name and the network application?

wenkokke:
Let's use # or @, which will look less like they have any business being in a variable name or numeric expression.

developing-ar:
Would this require changing the allowable variable names to include that character (currently only letters and digits are allowed)? 
I like `@`  as we can't use `#`  as it is not SMT-LIB compliant, and `-` is looks too similar to tensor indexing - unless that's not really a problem.

The main "objectives" are either that the network names are not numerically terminated (which might not be VNNLib) or manage to "encode" a separator - not sure if there's any-other option that using the special character.

MatthewDaggitt:
Yes, we'd have to make sure that whatever character we chose was allowed in the new VNNLib specification. I guess we could use `_` if we agreed with @Allenator1 that the parsers would split the variable correctly, e.g. for the input variable of type `Tensor [2,2]` for the 1st application of network `f`, then the VNNLib parser would correctly split `f1_X_0_1` as `f1_X`, `0`, `1`... that would mean we didn't have to add any extra characters to the spec.

developing-ar:
@MatthewDaggitt a little bit of a "hacky" idea - since just need to ensure that names should not end in a number, how about mapping each (present) digit to a letter i.e. 0 -> A, 1 -> B ... and then appending the index?
However, this could complicate the process of mapping counter-examples to the network.

wenkokke:
Isn't it just easier to append "__NNN"?

developing-ar:
> or the input variable of type Tensor [2,2] for the 1st application of network f, then the VNNLib parser would correctly split f1_X_0_1 as f1_X, 0, 1... that would mean we didn't have to add any extra characters to the spec.

Yes, but `_` is already used to separate the "list" of tensor indices from the name - so allowing `_` in the name would still require changing the spec. The current allowed format is as follows, where `letter+ digit*` is the variable name (generating this from the network name).

```
token TensorElement letter+ digit* '_' digit+ ('-' digit+)* ;
```
Unless @MatthewDaggitt meant that the parser allows the network index to "join" the tensor indices and splits it accordingly?

MatthewDaggitt:
Hmm the problem with letters is that it kind of complicates parsing the name for the reader.

How about generating names of the form `<network><appIndex><input/output>` so we'd have e.g. `f1input_0_0`. Doesn't require any special characters or `_`, relatively easy to read and the parsing is unambiguous?

> However, this could complicate the process of mapping counter-examples to the network.

We don't need to worry about this, Vehicle automatically stores the map of variable names in the query to variables internally in Vehicle. As long as the generated names are unique it should all work automatically.

developing-ar:
> How about generating names of the form <network><appIndex><input/output> so we'd have e.g. f1input_0_0. Doesn't require any special characters or _, relatively easy to read and the parsing is unambiguous?

Sounds good! Will still need to change the grammar (allow "any" combination of letters and numbers) but that's a reasonable fix since the original rule seems too constricting on variable names anyways.

There's still an issue of collisions if a network name ends in a number - can we do `<network><input/output><appIndex>`?

wenkokke:
It seems like a good idea to change the VNNlib spec to allow some special characters, but designate them as reserved for the use of tools compiling to VNNlib, i.e., you're allowed to use `@` or `#` in names, but you really shouldn't.

wenkokke:
Furthermore, there isn't actually a problem here as long as Verticale is more restrictive in what network names it accepts than the VNNlib spec, i.e., Vehicle can say "You're not allowed `@` in your network name, because that's for me to use." There's no reason Vehicle has to allow the same network names as the VNNlib spec.

developing-ar:
Please excuse the large PR - I've added tests for acasXu, mnist-robustness, windcontroller and monotonicity.

vehicle/tests/golden/compile/windController/VNNLib.queries/safe-query1.txt.golden
```diff
@@ -0,0 +1,14 @@
...
+; Metadata:
+;  - VNNLib query format version: unknown
+;  - Vehicle version: 0.16.1+dev
+
+(assert (<= (+ (- () (* 8.0 controller@0X_1)) controller@0Y_0) 2.75))
+(assert (<= () 7.25))
+(assert (>= () 0.75))
+(assert (<= () 7.25))
+(assert (>= () 0.75))
+(declare-network controller@0
```
MatthewDaggitt: The network declarations should be in front of the assertions?

vehicle/tests/golden/compile/acasXu/VNNLib.queries/property8-query3.txt.golden
```diff
@@ -0,0 +1,31 @@
+; WARNING: This file was generated automatically by Vehicle
+; and should not be modified manually!
+; Metadata:
+;  - VNNLib query format version: unknown
+;  - Vehicle version: 0.16.1+dev
+
+(assert (<= (- () acasXu@0Y_1) 0.0))
```
MatthewDaggitt: Minor but it would probably be nicer if we used the seperator character `@` between the index and the i/o status as well, e.g. `acasXu@0@Y_1`?

vehicle/tests/golden/compile/acasXu/VNNLib.queries/property8-query3.txt.golden
```diff
@@ -0,0 +1,31 @@
...
+(assert (<= (+ () acasXu@0Y_3) 0.0))
+(assert (<= () 40469.909))
```
MatthewDaggitt: I know this isn't your fault, but it looks like there is a bug here in the generation as well? The brackets should be a variable? Either in this PR or a separate PR could you take a look?

MatthewDaggitt (commented):
Otherwise looks great!

vehicle/tests/golden/compile/acasXu/VNNLib.queries/property8-query3.txt.golden
developing-ar: Will do - probably in a separate PR.

vehicle/tests/golden/compile/monotonicity/VNNLib.queries/monotonic-query1.txt.golden
```diff
@@ -0,0 +1,14 @@
...
+; and should not be modified manually!
+; Metadata:
+;  - VNNLib query format version: unknown
+;  - Vehicle version: 0.16.1+dev
+
+(assert (<= (- () f@0X_1) 0.0))
+(assert (< (+ () f@0Y_1) 0.0))
+(declare-network f@0
+  (declare-input f@0X Real 1)
+  (declare-output f@0Y Real 1))
```
developing-ar: Also seems to be another issue in the generation: out of scope indexing (1 for a size 1 tensor)

MatthewDaggitt: Hang on, that is the size of the tensor, not the tensor index... So 1 is correct?

developing-ar: `f@0X_1` shouldn't this start at 0? i.e. `f@0X_0`

MatthewDaggitt: Ah right you're referring to lines 7 and 8, rather than Lines 10 and 11. Yes, those should be indexed at `0` :+1:

vehicle/tests/golden/compile/acasXu/VNNLib.queries/property8-query3.txt.golden
developing-ar: PR: #893

MatthewDaggitt: Great, you should be able to merge/rebase once the other PR is in, and then let's merge this :+1:

MatthewDaggitt:
Given that these are all very minor issues, @developing-ar any chance we could get these fixed and merged in?

developing-ar:
@MatthewDaggitt yes, I'll try get it done asap

MatthewDaggitt:
Okay so there'll be a few conflicts with the large branch I've just landed. I'll resolve the merge conflicts tomorrow and then hopefully we can get many of the comments above sorted :crossed_fingers:

MatthewDaggitt:
Hi @developing-ar, just brought this up-to-date. My other refactoring has provided far more information in the form of `QueryVariableInfo` for you to play with when generating the VNNLib code. What needs to be done is as follows:

- [x] Move network declarations before assertions
- [ ] Switch to the `[i,j,k]` representation discussed on https://github.com/VNNLIB/VNNLIB-Standard/issues/63 
- [ ] Fix the off-by-1-indexing issue (though may be fixed by the previous point)
- [ ] Use underscores instead of `@`
- [ ] We also have enough information provided via `numberOfNetworkApps` in `QueryVariableInfo` to not have to include the network application number in the name if there is only a single network application!

MatthewDaggitt:
Not going to fix the last three tests as they will change again when the variable names are updated.

MatthewDaggitt:
If the tests pass, is this ready to go @developing-ar? :smile:

developing-ar:
@MatthewDaggitt Yes, it's good to go!

MatthewDaggitt (approved):

PR #891, opened 2025-04-26
Extend stable Python API - fixes and `list`

developing-ar:
# Description of Changes
- Fixes #883 (imports) to correctly export added functions and classes, and `check` option
- Added Python Bindings for `list`command - to `list_resources` and `list_properties`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,7 +12,6 @@
   - `export_to_solver`
   - `list_resources` and `list_properties` which output as JSON
 
-
 ## Version 0.16.1
 
 * Fixed detection of Marabou timeouts.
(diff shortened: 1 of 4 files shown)


wenkokke (approved):
Looks good to me!

MatthewDaggitt (approved):

MatthewDaggitt:
Before I merge, do you think you could add some entries in the `CHANGELOG.md` file for both this change and the adding of `--list` command line mode?

developing-ar:
I've added it under Version 0.17, or would it be under 0.16.2?

MatthewDaggitt:
Version 0.17 is good as we're changing functionality!

PR #895, opened 2025-05-17
Nix

faezs:
added a nix build for the haskell env that's mostly done but the python package requires some env definition to actually find all the libraries.


wenkokke:
Would you be willing to commit to maintaining this if it's merged?

Could you add a CI job that tests the Nix build?

Could you pin the versions of every piece of software that isn't otherwise specified, such as GHC, Cabal, Agda, the Agda standard library, and Python?

faezs:
> Would you be willing to commit to maintaining this if it's merged?

Yeah don't forsee not needing this any time soon :)

> Could you add a CI job that tests the Nix build?

I will add garnix support.

> Could you pin the versions of every piece of software that isn't otherwise specified, such as GHC, Cabal, Agda, the Agda standard library, and Python?

GHC+Cabal -> pinned at 9.8.4 for now but easy to swap out.
Python+ all the tensorflow jax ml stuff -> also pinned. 
more stolid verification can be found as soon as someone with an x86_64 generates and commits the lock file by cloning and running `nix run .#vp.lock`

I tried to do minimal diffs with the setup.py bootstrap though that all is handled with nix now.

Agda pinning needs to work. would make things easy if you could point to the commit/version of the agda that you're currently using and the commit of the stdlib as well

faezs:
The fun part now is getting marabou into the flake as well. We should also embed programs that other backends might need.

faezs:
@wenkokke could you please sign up for https://garnix.io and install its github app for this repo? It should automatically pick up the flake.nix on this branch and start the ci runs.

wenkokke:
> @wenkokke could you please sign up for https://garnix.io and install its github app for this repo? It should automatically pick up the flake.nix on this branch and start the ci runs.

I'd prefer to manage nix based runs with GHA only. Presumably, we could pretty easily set up a workflow that runs Nix build in a docker container?

faezs:
Is there a particular reason you prefer GHA? The problem with a GHA based
setup in my experience is usually that nix caching is a pain to configure.
Garnix has excellent caching and is free for public repos.

wenkokke:
Primarily the introduction of another technology that I'd have to manage and understand. I prefer GHA, since I could at least partially understand what the CI scripts are doing.

faezs:
> Primarily the introduction of another technology that I'd have to manage and understand. I prefer GHA, since I could at least partially understand what the CI scripts are doing.

Ah that's the best part about garnix! no CI scripts. by default it just builds everything in the flake outputs including devshells and checks etc. There are some config things we can do to incrementalize the builds if they end up taking a ridiculously long time but otherwise it's zero-config.

faezs:
I did find this though so if you're not convinced about garnix I'll integrate this

https://github.com/nix-community/nix-github-actions

MatthewDaggitt:
I'm going to close this @faezs but feel free to update it and reopen it if you are keen to add Nix support.

PR #896, opened 2025-05-18
Issue 892 : Validate enhancements

developing-ar:
# Description of changes

The `--counter-examples` and `--json` options for validate and with updated python bindings.

The `--counter-examples` flag prints all the assignments for properties from the provided verification cache.
The `--json` option outputs as the following:
```
{
    "status": "<verified|unverified|unknown>",
    "error": "<error-message>",                                    (only if status is unknown)
    "counter-examples": [...]                                      (only if --counter-examples)
}
```
Where, "counter-examples" is `[]` if there are no counter-examples, or a list of results objects.
For example, a result-object has the format:
```
{
      "property": "safe",
      "assignments": [
        {
          "\"x\"": "[ -2.4, -1.6 ]"
        },
        {
          "\"y\"": "[ -2.4, -1.6 ]"
        },
      ]
    }
```

### DIFF
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -501,7 +501,7 @@ outputCounterExamplesParser =
   switch $
     long "counter-examples"
       <> short 'x'
-      <> help "Output the counter-examples in the validate cache."
+      <> help "Output the counter-examples in the proof cache."
       <> internal
 
 propertyParser :: Parser [Text]
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -501,7 +501,7 @@ outputCounterExamplesParser =
   switch $
     long "counter-examples"
       <> short 'x'
-      <> help "Decode the counter-examples in the validate cache."
+      <> help "Output the counter-examples in the validate cache."
       <> internal
 
 propertyParser :: Parser [Text]
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -276,6 +276,8 @@ validateParser :: Parser ValidateOptions
 validateParser =
   ValidateOptions
     <$> validateCacheParser
+    <*> outputAsJSONParser
+    <*> outputCounterExamplesParser
 
 validateParserInfo :: ParserInfo ModeOptions
 validateParserInfo = info (Validate <$> validateParser) validateDescription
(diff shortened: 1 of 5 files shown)


developing-ar:
I have not added tests yet. Once/if the "format" has been finalized/approved I can add tests accordingly.

vehicle/src/Vehicle/CommandLine.hs
```diff
@@ -494,6 +496,14 @@ outputAsJSONParser =
       <> help "Output the program as JSON instead of text."
       <> internal
 
+outputCounterExamplesParser :: Parser Bool
+outputCounterExamplesParser =
+  switch $
+    long "counter-examples"
+      <> short 'x'
+      <> help "Output the counter-examples in the validate cache."
```
MatthewDaggitt: Can we change "validate cache" to "proof cache"

vehicle/src/Vehicle/Validate.hs
```diff
@@ -53,3 +90,36 @@ instance Pretty ValidateResult where
...
+
+instance ToJSON ValidateResult where
+  toJSON validateResult = case validateResult of
+    Verified -> object ["status" .= ("verified" :: String)]
+    Unverified -> object ["status" .= ("unverified" :: String)]
+    IntegrityError err ->
+      object
+        [ "status" .= ("unknown" :: String),
+          "error" .= (show $ pretty err :: String)
+        ]
```
MatthewDaggitt, line 104: Can we simply derive this JSON instance automatically rather than manually? If not, can we leave a comment explaining why not :)

vehicle/src/Vehicle/Validate.hs
```diff
@@ -53,3 +90,36 @@ instance Pretty ValidateResult where
...
+
+instance Pretty CounterExampleResult where
+  pretty (CounterExampleResult assignments propertyName)
+    | null assignments = mempty
+    | otherwise =
+        pretty propertyName
+          <> line
+          <> vsep (map (\(name, tensor) -> indent 2 $ pretty name <> ": " <> pretty tensor) assignments)
+
+instance ToJSON CounterExampleResult where
```
MatthewDaggitt, line 119: Again, can this be derived automatically?

MatthewDaggitt (commented):
Apart from those minor comments, looks great! Looking forward to seeing the tests.

vehicle/src/Vehicle/Validate.hs
developing-ar: It can, but it's not the easiest to parse. I've added a comment in [c894d71](https://github.com/vehicle-lang/vehicle/pull/896/commits/c894d71247da71deb4f87662bc107b1f1f0d3487):
```
-- | Won't derive generically so that the assignments "name" is the key,
-- | and to have the pretty list of assignments rather then the entire JSON object of a tensor.
```
Maybe the JSON of the whole tensor is actually better?
Then only the rational tensor JSON can be derived generically, as it's still easier "access" if the assignment name (for example `"x"`) is the key in the JSON.

vehicle/src/Vehicle/Validate.hs
developing-ar: Added comment in [c894d71](https://github.com/vehicle-lang/vehicle/pull/896/commits/c894d71247da71deb4f87662bc107b1f1f0d3487)

developing-ar:
As per discussion with @MatthewDaggitt, counter-examples will not be included as an option for validate. Only enhancement is for a JSON option which is in #897.

PR #897, opened 2025-05-22
Validate `--json` option

developing-ar:
# Description of changes
The `--json` option for `validate`  which is generically derived.

The JSON has the form for verified/unverified statuses:
```
{
  "tag": <"Verified"|"Unverified">
}
```
For error i.e. unknown status:
```
{
  "tag": "IntegrityError",
  "contents": {
    "tag": <"MissingResources"|"AlteredResources">,
    "contents": [
      {
        "name": string,
        "filePath": filepath string,
        "fileHash": int
      }, ...
    ]
  }
}
```


developing-ar:
I can add tests, but that might require a proof cache directory being added as well, with at least 1 showing an "error". Is there another way to go about this?

MatthewDaggitt:
Hmm, can you create a test using the "TestVerifier" executable that outputs a proof cache which you then run this command on? You should be able to run multiple commands in a row in the "run" field in the json test specifications.

e.g. see `TestVerify` test case in `WindController/test.json`

developing-ar:
@MatthewDaggitt I'm having an issue with the "recognition" of the cache directory for the tests. 
For example,
```
{
    "name": "ValidateJSON",
    "run": "vehicle validate -c verify-cache -j",
    "needs": ["verify-cache"]
  }
```
Errors with `Could not find needed files: "verify-cache"` even when the directory is present. 
I've tried `"verify-cache/*"` as well as "listing" all the files in the verify cache —`"verify-cache/.vcl-cache-index", ...` — which do not work. Do directories need to be specified (in "needs") differently?

To reproduce:
1. Add the above to the `windcontroller/test.json`
2. Create the cache with the `TestVerify` command appending `-c verify-cache`
3. Run `cabal test` for `windController`

MatthewDaggitt:
Thanks @developing-ar, I've just brought this up to date and made a minor tweak to the tests:
- moved them to the `validate` subfolder
- replaced the successful tests for `windController` and `acasXu` with a standalone success test. Those two files fail all the time for unrelated reasons, so best to decouple the validation tests so they don't rely so much on the rest of the vehicle compiler being correct.

I've set this to merge in when the CI passes :crossed_fingers:

PR #907, opened 2025-07-24
Add records to the Vehicle language

MatthewDaggitt:
Several envisaged uses:
1. (Short term) be able to declare differentiable logics within Vehicle itself.
2. (Long term) be able to coerce automatically between `Tensor` and records of tensors.
3. (Long term) support networks with multiple inputs and outputs.

@wenkokke I would really like your review on this as I know you have strong feelings about how we implement them. I've taken your advice and kept projections as special cases in the language rather than instantiate them as functions as Agda does.


wenkokke:
@MatthewDaggitt This is very cool!

MatthewDaggitt:
Yup, not sure what times you keep at the moment so I'd be free (UK times):
- 2am-2:30am
- 7am-9am
- 11-11:30am
- 3pm-4pm

Do any of those work?

MatthewDaggitt:
Conclusion of discussion: https://github.com/vehicle-lang/vehicle/issues/912

But merging in for now.

PR #927, opened 2025-08-09
Global `--json` option

developing-ar:
# Description of changes
Make the `--json` option global i.e. applicable for all vehicle `COMMAND`s.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,10 @@
 
 * Fixed various bugs that sometimes occured when compiling tensors with dimensions >= 2.
 
+### Command-line interface
+
+* Update the `--json` option flag to be globally applicable for Vehicle to output the result of implementing commands as machine-readable JSON. It is currently implemented in the commands: `list`, `validate`, and `compile` with the `--json` or `-j` flag.
+
 ## Version 0.18.0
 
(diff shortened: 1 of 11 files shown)


MatthewDaggitt (approved):

MatthewDaggitt:
Looks great, thanks! The only thing you need to do is to add a CHANGELOG explaining the changes to the command line format.

developing-ar:
@MatthewDaggitt is the description of the change sufficiently explanatory?

MatthewDaggitt:
So one problem is that you've edited the entry for a prior version which has already been released! You'll need to add a new entry to the top of the file under version ## Version 0.18.1.

The important thing in Changelog entries is to make it clear to the reader a) what has changed and b) whether that is breaking. For a) I would give an example of the old command and then an example of the new command. That should allow them to update their commands very easily.

developing-ar:
@MatthewDaggitt, my apologies! Yes - I've fixed that
I was also wondering if it would be worth recording the JSON schemas?

MatthewDaggitt:
> I was also wondering if it would be worth recording the JSON schemas?

No, I think it's fine for the moment. It's not going to be widely used feature, and they can always try it and find out.

PR #940, opened 2025-08-21
Handle tensor-level inequalities during query compilation

### LINKED ISSUE #938: Error: missing element in substitution 2
Another spec where we get miscalculate the substitution between internal variables and query variables:
```
type UnnormalisedImage = Tensor Real [1]

@network
classifier : UnnormalisedImage -> Real

@property
robust : Bool
robust = forall (x : UnnormalisedImage) . ((foreach c . -1) <= x) => classifier (foreach c . x ! c) > 0
```

MatthewDaggitt:
This is another refactor that kind of got away from me. 

In theory the change is very simple. Now that we have tensor level operations, we can also have tensor level inequalities. This adjusts the user variable elimination procedure to correctly handle these tensor level inequalities over variables.


MatthewDaggitt:
Fixes #938

PR #941, opened 2025-08-22
Typecheck json output

developing-ar:
# Description of Changes
Implemented the JSON output for type check errors by toJSON for `VehicleError`.

General error schema:
```
{
 JSON or str based on Error Type
}
```

### DIFF
--- a/vehicle/src/Vehicle/Compile/Error.hs
+++ b/vehicle/src/Vehicle/Compile/Error.hs
@@ -4,10 +4,14 @@ module Vehicle.Compile.Error where
 
 import Control.Exception (IOException)
 import Control.Monad.Except (MonadError, throwError)
+import Data.Aeson (ToJSON, object, toJSON, (.=))
 import Data.List.NonEmpty (NonEmpty)
 import Data.Text (Text)
 import Data.Typeable (Proxy)
 import Data.Void (Void)
+import GHC.Generics (Generic)
+import Prettyprinter (defaultLayoutOptions, layoutPretty)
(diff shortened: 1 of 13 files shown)


vehicle/src/Vehicle/Compile/Error.hs
```diff
@@ -195,5 +214,33 @@ instance Pretty VehicleError where
...
 
+data VehicleErrorJSON = VehicleErrorJSON
+  { errorType :: Text,
+    errorContent :: Aeson.Value
+  }
+  deriving (Generic, Show)
+
+instance ToJSON VehicleErrorJSON
+
+convertVehicleErrorToJSON :: VehicleError -> VehicleErrorJSON
```
MatthewDaggitt: I think this could probably be simplified. The user doesn't really care what the error type is, that is simply internal bookkeeping information. I think we could remove the `VehicleErrorJSON` class altogether and this function would simply have the type `VehicleError -> Aeson.Value`... 

Sound reasonable?

MatthewDaggitt (changes requested):
Otherwise looks great!

vehicle/src/Vehicle/Compile/Error.hs
developing-ar: Sounds good to me!

MatthewDaggitt:
@developing-ar I'm hoping to make a release tomorrow evening. Any chance we could get this merged in by then?

MatthewDaggitt (approved):

PR #950, opened 2025-09-18
List quantified variables

developing-ar:
List the bound Variables in a specification


vehicle/src/Vehicle/List.hs
```diff
@@ -57,11 +59,42 @@ printResources (Main decls) listEntities outputAsJSON = do
           else pretty listDecls
   programOutput outputDocs
 
+quantifiedVariableToListEntity :: Expr builtin -> Maybe ListEntity
+quantifiedVariableToListEntity = \case
+  BoundVar prov idx -> Just $ ListEntity {entitySort = "variable", entityName = pack (show idx), entityType = "quantified", entityProvenance = prov}
```
MatthewDaggitt: So this isn't quite right as we don't want every variable ever used by the user and this doesn't recurse all the way through the expression. We only want the tensor variables that are universally or existentially quantified over. So we need to 

1. specialise the `builtin` type to `Builtin` 
2. recurse all the way through (you can use the existing `traverseBuiltinsM` for this!)
3. search for things of the form `QuantifyRatTensor [(Lam binder body)]` and extract the name of the variable and the type from the binder.

We could also make this function much neater by having the type `MonadWriter [ListEntity] m => Expr builtin -> m ()`.

vehicle/src/Vehicle/List.hs
```diff
@@ -57,11 +59,42 @@ printResources (Main decls) listEntities outputAsJSON = do
...
+  _ -> Nothing
+
+parseGenericProgForQuantifiedVariable :: GenericDecl (Expr builtin) -> Maybe ListEntity
+parseGenericProgForQuantifiedVariable = \case
+  DefFunction _ _ _ _ expr -> quantifiedVariableToListEntity expr
+  DefAbstract _ _ _ expr -> quantifiedVariableToListEntity expr
+  DefRecord _ _ _ expr _ -> quantifiedVariableToListEntity expr
+
+listQuantifiedVariables :: [GenericDecl (Expr builtin)] -> [ListEntity]
+listQuantifiedVariables [] = []
```
MatthewDaggitt: Can we use `\case` and split on the cases on separate lines to avoid duplicating the function name more than we need to?

vehicle/src/Vehicle/List.hs
```diff
@@ -57,11 +59,42 @@ printResources (Main decls) listEntities outputAsJSON = do
...
+  let listDecls = listQuantifiedVariables decls
+  let outputDocs =
+        if outputAsJSON
+          then pretty $ unpack $ encodePretty' prettyJSONConfig $ toJSON listDecls
+          else pretty listDecls
+  programOutput outputDocs
+
 -- | Data Structure for listable entities
 data ListEntity = ListEntity
   { entitySort :: Text,
```
MatthewDaggitt: What's the difference between `entitySort` and `entityType`? Can we make them more structured than simply being `Text`?

MatthewDaggitt (changes requested):

vehicle/src/Vehicle/List.hs
developing-ar: `entityType` is the type that the entity has been declared with in the vehicle file.
`entitySort` is the "category" of the listable entity.
For example, when listing properties:
```bash
{
    "entityType": "Bool",
    "entitySort": "@property",
    "entityName": "safe"
}
```

I'm hoping for the quantified variables the output is something like:
```bash
{
    "entityType": "forall",                                    -- or "exists"
    "entitySort": "quantifiedVariable",                        -- or "quantified@variable"
    "entityName": "x",
    "entityProvenance" : ...
}
```

MatthewDaggitt: Ah right. Okay so I think we should probably do something like the following.

1. Change `ListableResources` into:
   ```haskell
   data ListableEntity
         = Network
         | Dataset
         | Parameter
         | Property
         | QuantifiedVariable
   ```
   and get rid of `ListableVariables` and `ListableEntities`

2. Change `ListEntity` to
    ```haskell
    data ListEntity = ListEntity
      { entity :: ListableEntity,
        entityName :: Text,
        entityProvenance :: Provenance
        entityType :: Text, -- TODO return a more structured representation of this type information 
      }
    ```

3. Maybe simplify the `vehicle list` command to simply print out all listable entities it finds rather than force the client to request each type one-by-one from Vehicle (re-type-checking each time). The consumer of the API can always filter the list themselves very simply to get e.g. the list of quantified variables.

Thoughts?

developing-ar: > Maybe simplify the vehicle list command to simply print out all listable entities it finds rather than force the client to request each type one-by-one from Vehicle (re-type-checking each time). The consumer of the API can always filter the list themselves very simply to get e.g. the list of quantified variables.

Wouldn't this make the output json quite large? How about having `vehicle list` by default print all the listable entities, and then passing flags like `--resources`, `--properties` etc. would filter the output json?

MatthewDaggitt: I mean large is relative, compared to some of the other JSON or files that Vehicle generates it will still be quite small and take milliseconds to parse. I'm also not sure we have a use case where a consumer would want some of them and not the others?

Having said that I'm happy with your proposed solution if you want to implement that!

developing-ar: That's true!
For example, I was thinking that listing properties might be done separately to "choose" the property to verify.

MatthewDaggitt: Except that the GUI will want to know the properties in advance at the same as the networks etc. to allow the user to choose the property to verify right?

developing-ar: You're right! I'll change the `list` command so it just lists all the listable entities

MatthewDaggitt:
Okay, so once again this turns out to be a little more complicated than I had hoped. When I started to think about it, to list all quantifiers it actually requires normalising the declaration.

Crucially, I have changed the JSON format to a) be more forwards compatible and b) to allow extra information to be provided on a per-entity basis. Namely:
- parameters now provide their inferability status
- properties now list all the quantified variables inside of them

MatthewDaggitt (approved):

developing-ar:
Looks like most of what I did had to change. @MatthewDaggitt thanks so much! 
> parameters now provide their inferability status

That makes sense on the vehicle side.

> properties now list all the quantified variables inside of them

It's so great to have the quantified variables linked to the property!
I'll add a separate PR to update the python bindings - in #965

PR #952, opened 2025-09-20
Upgrade BNFC to 2.9.6 to get rid of derive (Typeable)

### LINKED ISSUE #916: Remove deriving of Typeable
Deriving Typeable is a warning in the more recent versions of GHC and might become an error in the future.

MatthewDaggitt:
Fixes #916


MatthewDaggitt:
@wenkokke am I right in thinking that this is a problem inside your `manylinux/musllinux container`?

I'm want to use a snapshot of cabal in the last 6 months but it looks like the version of cabal inside your container is older than that, and therefore can't find it.

Do we need to update the cabal package index inside your container? If so, how do we go about doing it?

wenkokke:
> @wenkokke am I right in thinking that this is a problem inside your `manylinux/musllinux container`?
> 
> 
> 
> I'm want to use a snapshot of cabal in the last 6 months but it looks like the version of cabal inside your container is older than that, and therefore can't find it.
> 
> 
> 
> Do we need to update the cabal package index inside your container? If so, how do we go about doing it?

My containers are definitely older than that, but I think my Python build should definitely be calling cabal update.

wenkokke:
> > @wenkokke am I right in thinking that this is a problem inside your `manylinux/musllinux container`?
> > 
> > 
> > 
> > I'm want to use a snapshot of cabal in the last 6 months but it looks like the version of cabal inside your container is older than that, and therefore can't find it.
> > 
> > 
> > 
> > Do we need to update the cabal package index inside your container? If so, how do we go about doing it?
> 
> My containers are definitely older than that, but I think my Python build should definitely be calling cabal update.

As a workaround, you could try setting [`before-all`/`CIBW_BEFORE_ALL`](https://cibuildwheel.pypa.io/en/stable/options/#before-all) to run `cabal update` on Linux containers.

In practice, adding

```yaml
env:
  CIBW_BEFORE_ALL_LINUX: "cabal update"
```

to the CI job that runs cibuildwheel for Linux should solve this?

wenkokke:
Looks like it worked!

wenkokke (approved):

MatthewDaggitt:
Thanks for the suggestion, would have taken me an age to find that!

PR #953, opened 2025-09-20
Restrict fields of records annotated with @tensor

laurenpudz:
Restricts fields of records annotated with `@tensor` to fields which will allow the meaningful conversion of the record to a tensor.

Fields must be elements of tensors or tensors.
All fields must be of the same type.

Unit tests have been added for both of the above cases.


vehicle-syntax/src/Vehicle/Syntax/AST/Decl.hs
```diff
@@ -194,3 +194,6 @@ instance Pretty Annotation where
 
 isProperty :: [Annotation] -> Bool
 isProperty anns = AnnProperty `elem` anns
+
+isTensor :: [Annotation] -> Bool
+isTensor anns = AnnTensor `elem` anns
```
MatthewDaggitt: Hmm so the original `isProperty` wasn't a great name, and I'm not sure this is either as it's really just testing annotations, and the declaration definitely isn't a record! Could we refactor to both these names to `isAnnotatedAsProperty` and `isAnnotatedAsTensor`?

vehicle/src/Vehicle/Compile/Type/System.hs
```diff
@@ -39,6 +39,13 @@ class (Eq builtin, Hashable builtin, NormalisableBuiltin builtin, TypableBuiltin
     Type builtin ->
     m (Type builtin)
 
+  restrictRecordType ::
```
MatthewDaggitt: Technically we're restricting the fields not the type. Maybe `restrictRecordAnnotatedAsTensor`?

vehicle/src/Vehicle/Compile/Type/System.hs
```diff
@@ -39,6 +39,13 @@ class (Eq builtin, Hashable builtin, NormalisableBuiltin builtin, TypableBuiltin
...
+    (MonadTypeChecker builtin m) =>
+    DeclProvenance ->
+    Type builtin ->
+    [RecordField (Type builtin)] ->
+    m (Type builtin)
```
MatthewDaggitt: Also we don't actually edit the type ever right so this could just return `m ()`

vehicle/src/Vehicle/Data/Builtin/Standard/Instances.hs
```diff
@@ -177,7 +177,21 @@ allInstances =
                 tUnit,
             False
           ),
-          ----------------
+          -------------------------
+          -- ValidTensorLikeType --
+          -------------------------
+
```
MatthewDaggitt: Can we get rid of the extra line?

vehicle/src/Vehicle/Data/Builtin/Standard/Instances.hs
```diff
@@ -177,7 +177,21 @@ allInstances =
...
+          ( forAllTypes $ \t ->
+              forAllDims $ \ds ->
+                isTensorType t ds
```
MatthewDaggitt: I don't think we actually need this `isTensorType t ds` constraint? This will get generated and checked when the user writes `Tensor X Y`.

vehicle/tests/golden/error/tensorLike/nonMatchingRecordFields/TypeCheck.err.golden
```diff
@@ -0,0 +1,2 @@
+Error in file 'spec.vcl' at Line 4, Columns 8-13: . In particular 'Real' is not equal to 'Nat'.
+Fix: check your types
```
MatthewDaggitt, line 2: Yup before we merge this, we should create an issue to improve this error message. It should mention that this constraint is generated because of the @tensor annotation.

vehicle/tests/golden/error/tensorLike/invalidTensorLikeType/TypeCheck.err.golden
```diff
@@ -0,0 +1,6 @@
+Error in file 'spec.vcl' at Line 10, Columns 5-7: The type of @tensor 'f3':
+  R1
+is not supported. All fields of a record annotated with '@tensor' must have the same type and must be either:
```
MatthewDaggitt: Hmm this error message is actually actively wrong. See comment in error generation above.

vehicle/src/Vehicle/Compile/Print/TypingError.hs
```diff
@@ -251,12 +249,17 @@ typeRestrictionError ctx (TypeRestrictionOrigin freeEnv (ident, p) sort typ) _ca
             <+> "to a supported type"
     }
   where
+    supportedTypesDescription = case sort of
+      Left _ -> "The possible valid types for" <+> quotePretty sort <+> "annotated declarations are:"
+      Right RestrictedRecord -> "All fields of a record annotated with" <+> quotePretty sort <+> "must have the same type and must be either:"
```
MatthewDaggitt: To get a good error message here, I think you may need to case split earlier, and actually generate an entire fresh `problem` field depending on whether it is `Left` or `Right`.

vehicle/src/Vehicle/Compile/Print/TypingError.hs
```diff
@@ -251,12 +249,17 @@ typeRestrictionError ctx (TypeRestrictionOrigin freeEnv (ident, p) sort typ) _ca
...
-      RestrictedParameter Inferable -> [pretty NatType]
-      RestrictedParameter NonInferable -> map pretty [BoolType, IndexType, NatType, RatType]
-      RestrictedDataset -> ["List A    " <+> datasetElementTypes, "Vector A n" <+> datasetElementTypes]
-      RestrictedNetwork -> ["Tensor Rat [a_1, ..., a_n] -> Tensor Rat [b_1, ..., b_n]  (where 'a_i' and 'b_i' are all constants at compile time)"]
+      Left RestrictedProperty -> ["Bool", "Vector Bool n", "Tensor Bool ns"]
+      Left (RestrictedParameter Inferable) -> [pretty NatType]
+      Left (RestrictedParameter NonInferable) -> map pretty [BoolType, IndexType, NatType, RatType]
+      Left RestrictedDataset -> ["List A    " <+> datasetElementTypes, "Vector A n" <+> datasetElementTypes]
+      Left RestrictedNetwork -> ["Tensor Rat [a_1, ..., a_n] -> Tensor Rat [b_1, ..., b_n]  (where 'a_i' and 'b_i' are all constants at compile time)"]
+      Right RestrictedRecord -> ["The element of a tensor (e.g. Real), or", "A tensor (e.g. Tensor Real [...])"]
```
MatthewDaggitt: We should really generate all of these lists automatically from the instances we have in scope. Can you open an issue on the Issue tracker to note this?

vehicle/src/Vehicle/Data/Builtin/Standard/Type.hs
```diff
@@ -188,10 +191,29 @@ restrictStandardDeclType declSort (ident, p) typ = do
...
+  (MonadTypeChecker Builtin m) =>
+  DeclProvenance ->
+  Type Builtin ->
+  [RecordField (Type Builtin)] ->
+  m (Type Builtin)
+restrictStandardRecordType (_ident, _p) typ [] = return typ
+restrictStandardRecordType (ident, p) typ ((firstFieldName, firstFieldType) : restFields) = do
+  env <- getFreeEnv
+  let expr = BuiltinExpr p (TypeClass ValidTensorLikeType) [explicit firstFieldType]
+  let origin = InstanceTypeRestrictionOrigin $ TypeRestrictionOrigin env ((Identifier (modulePath ident) (nameOf firstFieldName)), provenanceOf firstFieldName) (Right RestrictedRecord) firstFieldType
```
MatthewDaggitt: This is a very long line. Could you split some of it out into intermediate variables?

vehicle/src/Vehicle/Data/Builtin/Standard/Type.hs
```diff
@@ -188,10 +191,29 @@ restrictStandardDeclType declSort (ident, p) typ = do
...
   return typ
 
+restrictStandardRecordType ::
+  forall m.
+  (MonadTypeChecker Builtin m) =>
+  DeclProvenance ->
+  Type Builtin ->
+  [RecordField (Type Builtin)] ->
+  m (Type Builtin)
+restrictStandardRecordType (_ident, _p) typ [] = return typ
```
MatthewDaggitt: Can we use a \case to pattern match on the last argument to avoid rewriting out the name `restrictStandardRecordType` more than necessary?

MatthewDaggitt (changes requested):
Looks really good! A bunch of minor comments, but the approach is right!

vehicle/src/Vehicle/Data/Builtin/Standard/Type.hs
```diff
@@ -188,10 +191,30 @@ restrictStandardDeclType declSort (ident, p) typ = do
...
 
+restrictStandardRecordAnnotatedAsTensorType ::
+  forall m.
+  (MonadTypeChecker Builtin m) =>
+  DeclProvenance ->
+  [RecordField (Type Builtin)] ->
+  m ()
+restrictStandardRecordAnnotatedAsTensorType (ident, p) fields = case fields of
+  [] -> return ()
+  ((firstFieldName, firstFieldType) : restFields) -> do
```
MatthewDaggitt: No need for the outer brackets

vehicle/src/Vehicle/Compile/Type.hs
```diff
@@ -189,8 +189,14 @@ typeCheckRecordDef p ident _anns uncheckedType uncheckedFields isUnused = do
     logCompilerSection2 MidDetail pass $
       traverse (checkRecordFieldDef ident) uncheckedFields
 
+  _ <-
+    if isAnnotatedAsTensor anns
```
MatthewDaggitt: This can now be simplified by using `when` (see other places in the compiler for how it works)

vehicle/src/Vehicle/Data/Builtin/Standard/Type.hs
```diff
@@ -188,10 +191,30 @@ restrictStandardDeclType declSort (ident, p) typ = do
...
+    env <- getFreeEnv
+    let expr = BuiltinExpr p (TypeClass ValidTensorLikeType) [explicit firstFieldType]
+    let firstFieldIdentifier = ((Identifier (modulePath ident) (nameOf firstFieldName)), provenanceOf firstFieldName)
+    let origin = InstanceTypeRestrictionOrigin $ TypeRestrictionOrigin env firstFieldIdentifier (Right RestrictedRecord) firstFieldType
+    _ <- createFreshInstanceConstraint False mempty p origin Irrelevant expr
+
+    -- Add unification constraints to make sure fields are all of the same type.
+    let fieldTypes = map snd restFields
+    _ <- traverse (createFreshUnificationConstraint p mempty (CheckingInstanceType origin) firstFieldType) fieldTypes
```
MatthewDaggitt: If we created a fresh `origin` object here, we could actually put the two field names we were checking here to improve the error message? We would need to refine `RestrictedRecord` into two different constructors, one for checking the first field is the right type, and one for checking pairs of types are equal?

MatthewDaggitt (commented):
A few more comments that I spotted after your refactorings.

MatthewDaggitt (approved):

PR #954, opened 2025-09-21
Verify json output

conanpodewitt:
**JSON output support and pipeline changes:**

* Added a `jsonOutput` field to the `VerifierSettings` data type and threaded an `OutputAsJSON` flag through the verification pipeline, including the `verify`, `verifyQueries`, and related functions. This allows toggling between human-readable and JSON output modes. [[1]](diffhunk://#diff-09ba3259be558f6bf6d35fa9b1db7f0d6e3ce0b376dfde48271ede03b3b6a48bL41-R46) [[2]](diffhunk://#diff-09ba3259be558f6bf6d35fa9b1db7f0d6e3ce0b376dfde48271ede03b3b6a48bL71-R89) [[3]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L295-R302)
* Updated the main verification functions (`verifySpecification`, `verifyProperty`, `verifyQuery`) to conditionally output either human-readable logs or JSON events based on the `jsonOutput` flag. In JSON mode, property summaries and per-query results are emitted as JSON objects. [[1]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L325-R353) [[2]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540R377-R404) [[3]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L446-R494)

**New JSON event types and serialisation:**

* Introduced new data types `QueryOutput` and `PropertySummary` to represent JSON events for per-query and per-property results. Implemented `ToJSON` instances for these types, customising field names for JSON output.
* Added necessary imports and language extensions (`DeriveGeneric`, `OverloadedStrings`, `ToJSON`, etc.) to support generic JSON encoding. [[1]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540R1-R3) [[2]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L21-R28) [[3]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540R40)

**Refactoring for context passing:**

* Updated the context passed through the verification monad transformers to include the total number of queries for each property, enabling richer JSON output (e.g., including `queryTotal` in `QueryOutput`). [[1]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L306-L313) [[2]](diffhunk://#diff-45f915ddd54a4d9b9a0cef9fee39232e6713bad63c9e39a73b63f448becaf540L446-R494)


vehicle/src/Vehicle/Verify.hs
```diff
@@ -68,24 +68,25 @@ compileAndVerifyQueries loggingSettings outputAsJSON VerifyOptions {..} verifyCo
...
 
--- | Tries to locate the executable for the verifier at the provided
--- location and falls back to the PATH variable if none provided. If not
--- found then the program will error.
-locateVerifierExecutable ::
-  (MonadIO m) =>
-  Verifier ->
-  Maybe VerifierExecutable ->
-  m VerifierExecutable
+locateVerifierExecutable :: (MonadIO m) => Verifier -> Maybe VerifierExecutable -> m VerifierExecutable
```
MatthewDaggitt: Is there a reason this has been changed? Can we keep the layout and the comment?

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -322,15 +328,29 @@ verifySpecification ::
   FilePath ->
   m ()
 verifySpecification verifierSettings queryFolder = do
-  programOutput "Verifying properties:"
+  let json = jsonOutput verifierSettings
+  unless (json) $ programOutput "Verifying properties:"
```
MatthewDaggitt: Don't need the brackets around `json` here and everywhere else.

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -322,15 +328,29 @@ verifySpecification ::
...
   let verificationPlanFile = specificationCacheIndexFileName queryFolder
   SpecificationCacheIndex {..} <- readSpecificationCacheIndex verificationPlanFile
   maybeIntegrityError <- checkIntegrityOfResources resourcesIntegrityInfo
   case maybeIntegrityError of
-    Just err -> programOutput $ "Resource error:" <+> pretty err
+    Just err -> unless (json) $ programOutput $ "Resource error:" <+> pretty err
     Nothing -> forM_ properties $ \(name, multiProperty) -> do
       stats <- execWriterT $ verifyMultiproperty verifierSettings queryFolder multiProperty
-      outputStats name stats
+      if json && numberVerified stats + numberFalsified stats + numberTimedOut stats + numberErrored stats > 1
```
MatthewDaggitt: I think we should put this in `outputStats`  rather than do it here?

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -648,3 +691,44 @@ createPropertyProgressBar (PropertyAddress _ name indices) numberOfQueries = do
 
 closePropertyProgressBar :: (MonadIO m, MonadStdIO m) => PropertyProgressBar -> m ()
 closePropertyProgressBar _progressBar = VIO.writeStdoutLn ""
+
+-- New JSON event types
+
+data QueryOutput = QueryOutput
```
MatthewDaggitt: Can we call this `QuerySummary` to reflect `PropertySummary`?

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -648,3 +691,44 @@ createPropertyProgressBar (PropertyAddress _ name indices) numberOfQueries = do
...
+    result :: String
+  }
+  deriving (Generic)
+
+instance ToJSON QueryOutput where
+  toJSON =
+    genericToJSON
+      defaultOptions
+        { fieldLabelModifier = \case
+            "qoType" -> "type"
```
MatthewDaggitt: This feels a bit unsatisfactory. Couldn't we just give `qoType` a better name and then we wouldn't have to provide a manual implementation of `toJSON` at all? Same comment for `PropertySummary` below.

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -354,28 +374,34 @@ verifyProperty verifierSettings verificationCache address = do
...
-        case errorOrResult of
-          Left err -> return $ PropertyErrored err
-          Right result -> return $ PropertyCompleted $ NonTrivial result
-
-  outputPropertyResult verifierSettings verificationCache address result
+      -- Log property header
+      logCompilerSection MinDetail ("Verifying property" <+> quotePretty address) (return ())
+      -- Determine number of queries and initialize progress bar
+      let numberOfQueries = propertySize queryMetaData
+      progressBar <-
```
MatthewDaggitt: Okay, so I see the problem and yes this is kind of gnarly to handle both! So I think what we need to do is to define an object as follows:
```haskell
data ProgressReporter
  { reportQuerySummary :: MonadIO m => QuerySummary -> m ()
  , reportPropertySummary :: MonadIO m => PropertySummary -> m ()
  , closeReporter :: MonadIO m => m () 
  }
```
and then create an appropriate instance of it depending on whether we are outputting JSON or printing a progress bar to the command line?

We then pass that instance through instead of the progress bar. Does that make sense? Welcome to advance Haskell :smile:

vehicle/src/Vehicle/Verify/Specification/IO.hs
```diff
@@ -354,28 +374,34 @@ verifyProperty verifierSettings verificationCache address = do
...
-        when (numberOfQueriesExecuted < numberOfQueries) $ do
-          closePropertyProgressBar progressBar
-
-        case errorOrResult of
-          Left err -> return $ PropertyErrored err
-          Right result -> return $ PropertyCompleted $ NonTrivial result
-
-  outputPropertyResult verifierSettings verificationCache address result
+      -- Log property header
+      logCompilerSection MinDetail ("Verifying property" <+> quotePretty address) (return ())
```
MatthewDaggitt: This change needs to reverted. You need the `do` at the end, and the stuff underneath to be indented as it was before, because this prints stuff both before and after what happens in the middle.

MatthewDaggitt (changes requested):
This is pretty amazing for a first try in Haskell, so great job @conanpodewitt! I have added some initial suggestions for how we can improve it. Let me know if more explanation would help.

vehicle/src/Vehicle/Verify/Specification/IO.hs
conanpodewitt: Hey @MatthewDaggitt I was wondering if you could give me some more pointers on this abstraction. I've been a little stuck on this.

vehicle/src/Vehicle/Verify.hs
conanpodewitt: Reverted back to:

```
locateVerifierExecutable ::
  (MonadIO m) =>
  Verifier ->
  Maybe VerifierExecutable ->
  m VerifierExecutable
```

vehicle/src/Vehicle/Verify/Specification/IO.hs
conanpodewitt: Fixed

vehicle/src/Vehicle/Verify/Specification/IO.hs
conanpodewitt: Now contained within `outputStats`.

vehicle/src/Vehicle/Verify/Specification/IO.hs
conanpodewitt: As requested: `data QuerySummary = QuerySummary`

vehicle/src/Vehicle/Verify/Specification/IO.hs
conanpodewitt: Replaced with:
```
-- JSON event types
data QuerySummary = QuerySummary
  { eventType :: String,
    propertyID :: Int,
    propertyName :: Aeson.Value,
    propertyIndices :: [Int],
    queryID :: Int,
    queryTotal :: Int,
    result :: String
  }
  deriving (Generic, Aeson.ToJSON)

data PropertySummary = PropertySummary
  { eventType :: String, 
    propertyName :: Aeson.Value,
    verified :: Int,
    falsified :: Int,
    timedOut :: Int,
    errored :: Int
  }
  deriving (Generic, Aeson.ToJSON)
```

vehicle/src/Vehicle/Verify/Specification/IO.hs
MatthewDaggitt: When writing some hints I've realised this is actually quite fiddly, so I'm going to have a go at this myself now!

conanpodewitt: I was kind of feeling that, i'll move on to other issues in the meantime, but please let me know what I can do to help

MatthewDaggitt:
@conanpodewitt, @developing-ar, @Allenator1, I've got it so it outputs JSON. 

Can you have a look at `vehicle/tests/golden/compile/windController/TestVerifyJSON.out.golden` in this PR and check that it has all the information you want? If it does, then I'll merge this in. The datatypes that the JSON are generated from are at the bottom of `Reporting.hs`

vehicle/tests/golden/compile/windController/TestVerifyJSON.out.golden
```diff
@@ -0,0 +1,62 @@
+{
+  "tag": "VerificationStart"
```
MatthewDaggitt, line 2: Generated JSON is here.

Allenator1: For MultiPropertyStart, it would be good to have some information about the property. E.g., is it a for-all or exists property. Additionally, QueryFinish might have the "negated" parameter. This way the user can easily know how the query contributes to the property without relying on the vcl-plan or waiting till the finish. 

A very small point is how the JSON is produced for the end of the verification. The way we handle verification output is by having a callback function on received chunks from vehicle_lang.verify. If verification produces multiple top-level JSON objects, its slightly trickier to decode (we'll have to decode it line by line rather than using json.loads).

MatthewDaggitt: > it would be good to have some information about the property. E.g., is it a for-all or exists property.

This unfortunately doesn't make sense at either the `MultiProperty` or even `Property` level, e.g. `(forall x. f x > 0) and (exists x . f x < 1)` is a valid property that has both an exists and forall in. I can give this at the query level if it would be useful?

> The way we handle verification output is by having a callback function on received chunks from vehicle_lang.verify.

So the idea here is that Vehicle will produces a new top-level JSON object for each event, e.g. after it finishes processing each query. Do you want it to produce all the JSON at once after verification is finished? I assume I'm misunderstanding as that doesn't seem like we could get a responsive UI out of that...

Can you give me a brief example of the format you would prefer?

MatthewDaggitt: Oh right, you want 
```
{
      "propertyName": "safe",
      "propertyIndices": [],
      "numberOfQueries": 2
    },
    {
      "verified": false,
      "erroredQueryID": null,
      "errorMessage": null
    }
```
to be
```
{
      "propertyName": "safe",
      "propertyIndices": [],
      "numberOfQueries": 2,
      "verified": false,
      "erroredQueryID": null,
      "errorMessage": null
    }
```
However, these aren't top-level JSON objects so it should be possible to load in:
```
{
  "tag": "PropertyFinish",
  "contents": [
    {
      "propertyName": "safe",
      "propertyIndices": [],
      "numberOfQueries": 2
    },
    {
      "verified": false,
      "erroredQueryID": null,
      "errorMessage": null
    }
  ]
}
```
as one JSON chunk?

MatthewDaggitt: The latter format seemed better to me, as then you can compare the first JSON object for equality when pairing up start and finish events?

Allenator1: I like the the latter format better as well, since it nicely seperates the property information from its verification status. My question was whether JSON like:
```
{
  "tag": "MultiPropertyFinish",
  "contents": {
    "propertyName": "safe"
  }
}
```
would be outputted at the same time as:
```
{
  "tag": "VerificationFinish"
}
```
Since they are produced by vehicle as seperate strings as you mentioned, there shouldn't be an issue as they would be processed as seperate callbacks by the GUI.

Regarding your first point, I forgot again that properties can be complex trees! Unfortunately I don't think storing this information at the query level would solve this issue, so we will have to rely on the vcl-plan.

MatthewDaggitt: Okay so does this need any tweaking? :smile:

Allenator1: Nope you're good! Sorry that wasn't clear

MatthewDaggitt (approved):

PR #965, opened 2025-10-02
List quantified variables python

developing-ar:
Updating the python bindings for the `list` command.
- DEPRECIATED `list_resources`, and `list_properties` to use `list`
- Updated changelog
- FIX(?) remove outer double quotes from the filenames of saved assignments in the cache. For example, the assignment for quantified variable `x` was saved as `"x"`.

**Note: This should be merged after #950.**

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,17 +4,13 @@
 
 ### Command-line interface
 
-* (BREAKING) For `list` the `properties` and `resources` subcommand are depreciated. The `list` command only outputs as JSON including changes to the JSON schema.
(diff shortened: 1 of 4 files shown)


developing-ar:
@MatthewDaggitt I've put the changes under version **0.21.0** instead as there are a few breaking changes - would the next release have to be version **0.20.1**?

MatthewDaggitt (approved):

MatthewDaggitt:
Yes, good shout. I will make the release as v0.21.0

PR #989, opened 2025-11-09
Updated 'vehicle compile' command-line to have submodes

MatthewDaggitt:


MatthewDaggitt:
This is needed as once we have custom loss logics and arbitrary verifiers then we will no longer be able to distinguish via the `target` option whether we are compiling to loss, queries or ITPs.

PR #992, opened 2025-11-11
Generate function to produce the tensor version of a record 

laurenpudz:


MatthewDaggitt:
Right got it working! Couple of changes:

- Split the scope checking module up as it was getting way too big. No major changes, your code has just moved to [vehicle/src/Vehicle/Compile/Scope/RecordInstances.hs](https://github.com/vehicle-lang/vehicle/pull/992/files#diff-5c25e72afc579c7c64c365f80cc679524869e9dceaec57ad94ebb0ef7117497a)
- The only bug was extracting the element and dimensions type from the field. However, as the comment I added shows, we can't actually know the element and dimensions type of the fields at scope checking time as the user may be using an alias for the tensor... However, the type-checker can always infer them so we can actually just leave them as holes!

I think it should work now and should be easy enough to extend with the `tensorToX` construction.

MatthewDaggitt:
@laurenpudz I'm doing some unrelated work on the scope checker so I'm going to merge this in. If there are any error messages that need improving, please could you do open a fresh PR to finish them.

PR #1000, opened 2025-11-24
Test and improve error messages for zero field records annotated with @tensor

laurenpudz:
- Constructor of the error thrown in scope checking when finding a zero field record with a `@tensor` annotation has been renamed from `UnsupportedTensorAnnotation` to `ZeroFieldTensorLike` to be more specific and more similar to names for unsupported networks, paramaters etc...
- Have moved the constructor into its own section in the `CompileError` type (as it was previously in `Other`)
- Added detail to the error message. It should now read:
```
Error in file <file> at <provinance>: cannot use zero field record <ident> as a tensor.
Fix: remove the @tensor annotation or add fields to the record.
```

- Added test to test that error is thrown when a record with no fields is annotated with `@tensor`


vehicle/src/Vehicle/Compile/Print/Error.hs
```diff
@@ -640,6 +634,16 @@ formatCompileError = \case
               <+> "be used as the dimension of a dataset"
               <+> "(networks will be supported later)."
       }
+  -- TensorLike errors
+  ZeroFieldTensorLike (ident, p) ->
+    VehicleError
+      { provenance = Just p,
+        problem =
+          "cannot use zero field record"
```
MatthewDaggitt: Maybe "annotating a record with zero fields with `@tensor` is not currrently supported"?

MatthewDaggitt (changes requested):
Otherwise looks great!

MatthewDaggitt:
The test failure looks spurious. I hope when you push again, the problem will fix itself :crossed_fingers:

vehicle/src/Vehicle/Compile/Print/Error.hs
laurenpudz: Can do! Is it ok that we won't mention the name of the problematic record type anywhere in the error message?

MatthewDaggitt: Ah nice spot! How about "annotating the record X with `@tensor` is not currently supported as it has no fields"?

MatthewDaggitt (approved):

PR #1001, opened 2025-11-24
Allow modifications to instance database in typechecker state

laurenpudz:
Removes instance database from `ReaderT` in `TypeCheckerTInternals` and adds it as a field in `TypeCheckerState`.
This means we can now modify the instance database using `modifyTypeCheckerState`.

Also adds the instance annotation `@instance` and adds this to function definitions generated for record to tensor conversion.


vehicle/src/Vehicle/Compile/Scope/RecordInstances.hs
```diff
@@ -71,4 +71,4 @@ createRecordToTensor p recordIdent fieldElementType fieldDimensions fields = do
         let tensorElements = fmap (\(fieldName, _) -> recordAcc record (recordIdent, fieldName)) fields
         stackTensor fieldElementType firstDimension fieldDimensions tensorElements
 
-  DefFunction p functionIdent mempty functionType functionBody
+  DefFunction p functionIdent [AnnInstance] functionType functionBody
```
MatthewDaggitt: I wouldn't make this explicitly the instance in this PR, otherwise we will have to change the function type to the type class (which doesn't yet exist!)

vehicle/src/Vehicle/Compile/Type/Monad/Class.hs
```diff
@@ -109,49 +111,42 @@ class (MonadCompile m, MonadFreeContext builtin m, NormalisableBuiltin builtin,
   modifyTypeCheckerState :: (TypeCheckerState builtin -> TypeCheckerState builtin) -> m ()
   getFreshName :: Type builtin -> m Name
   clearFreshNames :: Proxy builtin -> m ()
-  getInstanceCandidates :: m (InstanceDatabase builtin)
```
MatthewDaggitt, line 112: Removing this is really nice, but could we make it a derived method instead in the `Operations` section of this file? That would minimise the changes needed upstream!

vehicle/src/Vehicle/Compile/Type/Monad/Class.hs
```diff
@@ -365,7 +360,8 @@ prettyMetaInternal m t = pretty m <+> ":" <+> prettyVerbose t
 clearMetaCtx :: forall builtin m. (MonadTypeChecker builtin m) => Proxy builtin -> m ()
 clearMetaCtx _ = do
   logDebug MaxDetail "Clearing meta-variable context"
-  modifyTypeCheckerState @builtin $ const emptyTypeCheckerState
+  instanceDatabase <- instanceCandidates <$> getTypeCheckerState @builtin
+  modifyTypeCheckerState @builtin $ const emptyTypeCheckerState {instanceCandidates = instanceDatabase}
```
MatthewDaggitt: Could probably write this more efficiently as: 
```haskell
modifyTypeCheckerState @builtin $ \state -> 
   emptyTypeCheckerState {instanceCandidates = instanceCandidates state}
```
(haven't type checked this snippet). That would avoid us getting and then modifying the state!

vehicle/src/Vehicle/Compile/Type/Monad.hs
```diff
@@ -62,13 +63,14 @@ import Vehicle.Data.Variable.Bound.Context.Generic
...
-  (Monad m) =>
+  (Monad m, Hashable builtin) =>
   FreeCtx builtin ->
   InstanceDatabase builtin ->
   TypeCheckerT builtin m a ->
   m a
-runTypeCheckerTInitially freeCtx instanceCandidates e =
-  fst <$> runTypeCheckerT freeCtx instanceCandidates emptyTypeCheckerState e
+runTypeCheckerTInitially freeCtx instanceDatabase e = do
+  let state = (\s -> s {instanceCandidates = instanceDatabase}) emptyTypeCheckerState
```
MatthewDaggitt: Can write this as:
```haskell
emptyTypeCheckerState { instanceCandidates = instanceDatabase } 
```
In fact it would make more sense for the `emptyTypeCheckerState` function to take the instance database as a direct argument?

vehicle/src/Vehicle/Compile/Type/Constraint/InstanceSolver.hs
```diff
@@ -66,7 +64,7 @@ solveInstanceConstraint depth constraint = do
   logDebug MaxDetail $ "Forced:" <+> prettyExternal normConstraint
 
   let goal = instanceGoal $ objectIn normConstraint
-  database <- getInstanceCandidates
+  database <- instanceCandidates <$> getTypeCheckerState @builtin
```
MatthewDaggitt: This could stay `getInstanceCandidates` if earlier suggestion was implemented.

MatthewDaggitt (changes requested):
Looks great. A few minor suggestions.

MatthewDaggitt (approved):

PR #1003, opened 2025-11-27
ggustavs/feat/python update ast

ggustavs:
Finished updating the python API parsing of the JSON AST produced by the Vehicle compiler.
Added pytorch as a loss backend
Added default samplers for both backends to allow out of the box usage for training. These can be overwritten if a user wants to provide their own sampler. (look at load_specification for more details in compile.__init__.py)


vehicle-python/.python-version
```diff
@@ -1,5 +1 @@
-3.9
 3.10
```
MatthewDaggitt, line 1: I know we talked about dropping 3.9. We don't want to drop all the later ones right?

vehicle-python/pyproject.toml
```diff
@@ -15,8 +15,15 @@ readme = 'README.md'
 license = 'MIT'
 license-files = [ "LICENSE" ]
 dynamic = ["version"]
-requires-python = ">=3.9,<3.14"
-dependencies = ["typing_extensions >=4.6,<5", "jaxtyping>=0.2, <0.3"]
+requires-python = ">=3.10,<3.14"
+dependencies = [
+    "typing-extensions>=4.6,<5",
+    "jaxtyping>=0.2, <0.3",
+    "tensorflow>=2.20.0",
```
MatthewDaggitt: Shouldn't `tensorflow` and `torch` be part of `[project.optional-dependencies]` dependencies rather than `[project]`. We don't want to force users to install both right?

vehicle-python/pyproject.toml
```diff
@@ -25,7 +32,10 @@ pygments = ["pygments >=2.14,<3"]
 numpy = ["numpy >=1.21,<3"]
 jax = ["jax >=0.4.26"]
 tensorflow = [
-  "tensorflow >=2,<3; (python_version>='3.9' and python_version<'3.13') and ((sys_platform=='linux' and (platform_machine=='x86_64' or platform_machine=='aarch64')) or (sys_platform=='darwin' and platform_machine=='x86_64') or (sys_platform=='win32' and platform_machine=='AMD64'))"
+  "tensorflow >=2,<3; (python_version>='3.10' and python_version<'3.13') and ((sys_platform=='linux' and (platform_machine=='x86_64' or platform_machine=='aarch64')) or (sys_platform=='darwin' and platform_machine=='x86_64') or (sys_platform=='win32' and platform_machine=='AMD64'))"
+]
+pytorch = [
+  "torch >=2.0,<3.0; python_version>='3.10' and python_version<'3.13'"
```
MatthewDaggitt, line 36: @wenkokke do we really need these python version bounds on tensorflow and torch if we are already restricting what versions of python we are using? If we do, is there a way we can take the lower bound and upper bound automatically rather than having to maintain them here?

vehicle-python/src/vehicle_lang/compile/_abc/_builtins.py
```diff
@@ -0,0 +1,140 @@
...
+    def ConstTensor(self, value: vcl.Rat, shape: Sequence[vcl.Index]) -> vcl.Tensor: ...
+
+    @abstractmethod
+    def DenseTensor(
+        self, values: Sequence[vcl.Rat], shape: Sequence[vcl.Index]
+    ) -> vcl.Tensor: ...
+
+
+@dataclass(frozen=True, init=False)
+class ABCBuiltins(
```
MatthewDaggitt: Could we have a little bit of documentation about what this class is for and what the difference is between this and the `Builtins` class above? It seems to be a mix of both lists and tensors, but we have `DenseTensor` in the class above?

vehicle-python/src/vehicle_lang/compile/_abc/_builtins.py
```diff
@@ -0,0 +1,140 @@
...
+from dataclasses import dataclass
+from typing import Any, Generic, Sequence
+
+from typing_extensions import TypeAlias, TypeVar, override
+
+from .._ast._nodes import Tensor
+from . import _types as vcl
+
+_S = TypeVar("_S")
+_T = TypeVar("_T")
```
MatthewDaggitt: These don't seem to be used in this file?

vehicle-python/src/vehicle_lang/compile/_abc/_samplers.py
```diff
@@ -0,0 +1,36 @@
...
+        Calculates the loss based on the provided bounds and search lambda.
+
+        Args:
+            dims: The dimensions for the sampling.
+            lower_bound: The lower bound tensor.
+            upper_bound: The upper bound tensor.
+            search_lambda: A callable representing the search lambda.
+            minimise: A flag indicating whether to minimise the search_lambda.
+        Returns:
+            Sequence[vcl.Tensor]: The computed loss. If the size is greater than 1,
```
MatthewDaggitt: Can we document that this should be a 1D tensor? And that it must lie between `lower_bound` and `upper_bound`? Does this documentation transfer automatically to its subclasses?

vehicle-python/src/vehicle_lang/compile/_ast/_decode.py
```diff
@@ -12,16 +12,15 @@
...
     Type,
-    Union,
     cast,
 )
 
 from typing_extensions import (
     Literal,
     TypeAlias,
     TypeVar,
+    Union,
```
MatthewDaggitt, line 23: Do we need this import? I thought the point of upgrading the Python version was that we didn't?

vehicle-python/uv.lock
```diff
@@ -0,0 +1,1795 @@
+version = 1
```
MatthewDaggitt, line 1: Does this file need to be committed?

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @override
+    def MaxRatTensor(self, x: tf.Tensor, y: tf.Tensor) -> tf.Tensor:
+        return tf.maximum(x, y)
+
+    @override
+    def ReduceAddRatTensor(
+        self, e: float, xs: tf.Tensor | Sequence[tf.Tensor]
+    ) -> tf.Tensor:
+        # e is the identity element (0 for addition), xs is the samples to reduce
```
MatthewDaggitt: This implementation looks weird to me. Shouldn't the definition of this whole function simply be `tf.add(tf.reduce_sum(xs), e)`? Same for all the reduction operations.

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
```diff
@@ -0,0 +1,152 @@
...
+    ) -> tf.Tensor:
+        # Despite the name, this implements element indexing (At operator in Haskell)
+        # The JSON AST uses 'DimensionLookup' but semantics are element access
+
+        # Handle tuple/sequence case (from StackTensor or similar)
+        if isinstance(xs, (tuple, list)):
+            return xs[i]
+
+        # Handle scalar tensor case - can't be indexed
+        if xs.shape.ndims == 0:
```
MatthewDaggitt, line 122: You don't need to handle this case. The Vehicle type-checker should catch this. If you do want to check for it then we should throw an error here rather than silently ignore? Same comment for torch backend I assume?

vehicle-python/src/vehicle_lang/compile/pytorch/_builtins.py
```diff
@@ -0,0 +1,155 @@
...
+        # Despite the name, this implements element indexing (At operator in Haskell)
+        # The JSON AST uses 'DimensionLookup' but semantics are element access
+
+        # Handle tuple/sequence case (from StackTensor or similar)
+        if isinstance(xs, (tuple, list)):
+            return xs[i]
+
+        # Handle scalar tensor case - can't be indexed
+        if xs.ndim == 0:
+            return xs
```
MatthewDaggitt: Same comment, not necessary?

vehicle-python/src/vehicle_lang/compile/pytorch/_builtins.py
```diff
@@ -0,0 +1,155 @@
...
+    def MaxRatTensor(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        return torch.maximum(torch.as_tensor(x), torch.as_tensor(y))
+
+    @override
+    def ReduceAddRatTensor(
+        self, e: float, xs: torch.Tensor | Sequence[torch.Tensor]
+    ) -> torch.Tensor:
+        # e is the identity element (0 for addition), xs is the samples to reduce
+        xs = torch.stack(list(xs))
+        return torch.sum(xs)
```
MatthewDaggitt: Same comment as Tensorflow?

vehicle-python/src/vehicle_lang/compile/pytorch/_builtins.py
```diff
@@ -0,0 +1,155 @@
...
+    dtype_index: torch.dtype = torch.int32
+    dtype_rat: torch.dtype = torch.float32
+
+    @override
+    def Index(self, value: int) -> int:
+        return value
+
+    @override
+    def RatTensor(self, value: _nodes.Tensor) -> torch.Tensor:
+        match value.value:
```
MatthewDaggitt, line 44: Hmm where does this `Fraction` class come from? When will we be receiving a single rational value? The compiler shouldn't be returning a single value here, it should only return tensors...

vehicle-python/src/vehicle_lang/compile/python/_translation.py
```diff
@@ -0,0 +1,544 @@
...
+        return py.Lambda(
+            args=py_binder(self.translate_binder(expression.binder)),
+            body=self.translate_expression(expression.body),
+            **asdict(vcl.MISSING),
+        )
+
+    def translate_Pi(self, expression: vcl.Pi) -> py.expr:
+        raise EraseType()
+
+    def translate_PartialApp(self, expression: vcl.PartialApp) -> py.expr:
```
MatthewDaggitt: I don't think this and therefore `py_partial_app` is needed any more?

examples/simple.vcl
```diff
@@ -0,0 +1,11 @@
+pi = 3.14159
```
MatthewDaggitt: Should this have been committed? Probably shouldn't be here even if it was meant to be included?

vehicle-python/tests/test_pytorch_integration.py
```diff
@@ -0,0 +1,160 @@
...
+            "compile",
+            "loss",
+            "--logic",
+            "DL2Loss",
+            f"--specification={spec_path}",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=60,
+        cwd="/home/gus/University/vehicle/vehicle-python",
```
MatthewDaggitt: Lots of hard-coded paths in this file?

vehicle-python/tests/test_pytorch_backend.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_pytorch_builtins_advanced_operations() -> None:
+    """Test advanced PyTorch builtin operations."""
+    from vehicle_lang.compile.pytorch._builtins import PyTorchBuiltins
+
+    builtins = PyTorchBuiltins()
+
+    # Test reductions
+    x = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
+    sum_result = builtins.ReduceAddRatTensor(0, x)
```
MatthewDaggitt, line 33: I think this test will fail if you put a value in other than `0`?

vehicle-python/tests/test_pytorch_backend.py
```diff
@@ -0,0 +1,85 @@
...
+    """Test that PyTorchTranslation can be instantiated."""
+    from vehicle_lang.compile.pytorch._translation import PyTorchTranslation
+
+    translation = PyTorchTranslation()
+    assert translation is not None
+    assert hasattr(translation, "builtins")
+    assert hasattr(translation.builtins, "AddRatTensor")
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")  # type: ignore[misc]
```
MatthewDaggitt, line 58: Will this test run on CI?

vehicle-python/src/vehicle_lang/compile/_ast/_nodes.py
```diff
@@ -0,0 +1,385 @@
...
+
+    input_type: BuiltinType
+    output_type: BuiltinType
+
+
+@dataclass(frozen=True)
+class RatType(BuiltinType):
+    """Rational number type: RatType"""
+
+    # RatType has no contents in JSON - it's just a tag
```
MatthewDaggitt: Do we need to document this? We don't document for the other empty types?

vehicle-python/tests/test_lossdl2_load.py
```diff
@@ -15,4 +14,8 @@
 )  # type: ignore[misc]
 def test_lossdl2_load(specification_path: Path) -> None:
     print(f"Load {specification_path}")
+    if specification_path == GOLDEN_PATH / "test_quantifier_all.vcl":
```
MatthewDaggitt, line 17: Can we open an issue about this on Github with the pointers of what needs to change when this is fixed?

MatthewDaggitt (changes requested):
This looks amazing @ggustavs

I have a bunch of minor comments but nothing major.

vehicle-python/.python-version
ggustavs: yeah, and we are not dropping them. The .python-version file essentially just specifies what version should be used during development.

vehicle-python/pyproject.toml
ggustavs: That's true, I'll move them out. Forgot to double-check dependencies :D

vehicle-python/src/vehicle_lang/compile/_abc/_builtins.py
ggustavs: No difference. I had it there because that's how it was in the original code, but now that I actually went through it, Builtins was entirely unused. Switched to just ABCBuiltins now :)

vehicle-python/src/vehicle_lang/compile/_abc/_builtins.py
ggustavs: Yup, removed them

vehicle-python/tests/test_pytorch_integration.py
ggustavs: Oops, fixed with proper paths

vehicle-python/tests/test_pytorch_backend.py
ggustavs: I might not have understood what e is supposed to represent in all of the reduction nodes then, because I have been treating it as the identity element, so I ignore it  and just applying the relevant operation for each node to save on that one operation that does not contribute. 
Is it something else?

vehicle-python/tests/test_pytorch_backend.py
ggustavs: If it has CUDA, otherwise no. It's only relevant for checking compatibility with CUDA.

vehicle-python/src/vehicle_lang/compile/_ast/_nodes.py
ggustavs: Removed, I probably made that comment while still figuring out the AST :D

examples/simple.vcl
ggustavs: Yeah that's a mistake, left over from initial testing of compilation

vehicle-python/src/vehicle_lang/compile/_ast/_decode.py
ggustavs: No, I just haven't touched this file except for replacing Tuple for tuple, etc. in the entire codebase. I'll clean it up :)

vehicle-python/uv.lock
ggustavs: I mean for supporting UV and making dependency management easier in the future, yes. It's essentially a very detailed requirements.txt file ([docs](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile))

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
ggustavs: Alrighty, Ill just raise an exception here

vehicle-python/src/vehicle_lang/compile/_abc/_samplers.py
ggustavs: It does automatically get passed to subclasses if there isnt a local docstring

vehicle-python/src/vehicle_lang/compile/_ast/_decode.py
ggustavs: Nevermind I did look at this. We need it for the later registration of UnionDecoder

vehicle-python/tests/test_pytorch_backend.py
ggustavs: Fixed it as per your comments :)

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
ggustavs: image
Seems I added it because of specifications like these

vehicle-python/src/vehicle_lang/compile/_abc/_samplers.py
ggustavs: Wait, the sampler is supposed to return a loss. How are the bounds relevant to the loss? But yes, we can document that the internally generated samples should be between `lower_bound` and `upper_bound`

vehicle-python/pyproject.toml
wenkokke: They were there when those were not the same bounds we supported. Whenever that's the case, we need them. They ensure that the tests run correctly for versions that don't support those packages.

vehicle-python/uv.lock
wenkokke: Yes, it should be committed.

vehicle-python/src/vehicle_lang/compile/python/_translation.py
wenkokke: It shouldn't, no.

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
ggustavs: I changed the specification now and it runs fine, but I would like to know exactly what is the issue here (python side or the compiler itself)

vehicle-python/src/vehicle_lang/compile/pytorch/_builtins.py
ggustavs: It's in the case of a ConstantTensor. No need to return a full tensor when you can return the value and shape.
Reminder that we are checking value.value, not value itself. This follows the definition of a ConstantTensor: (ConstantTensor TensorShape a)

vehicle-python/src/vehicle_lang/compile/python/_translation.py
ggustavs: Why not? They still are in JSON.hs

ggustavs: ooooh you are talking about translate_PartialApp, not translate_Pi

vehicle-python/tests/test_lossdl2_load.py
ggustavs: https://github.com/vehicle-lang/vehicle/issues/1004

vehicle-python/src/vehicle_lang/compile/python/_translation.py
ggustavs: Removed partial app

vehicle-python/src/vehicle_lang/compile/_abc/_samplers.py
MatthewDaggitt: Ah sorry, my mistake. Just me having a brain fade moment. Yes of course, that is internal to the sampler.

vehicle-python/src/vehicle_lang/compile/tensorflow/_builtins.py
MatthewDaggitt: That looks like a bug in the compiler then, as the output of the network should be 1-dimensional. Can you open an issue with the original specification and I'll try and have a look at it?

MatthewDaggitt (approved):

PR #1012, opened 2025-12-02
Isabelle Backend for Vehicle

samysweb:
I thought I'd open a pull request so that people can look at the current state more easily if interested.
This is still work in progress but has come along pretty far now.

**Remaining ToDos:**
- [x] Compilation for records (Isabelle has records, so this should be straightforward)
- [x] Refactor proof for Wind Controller (there were some changes after completing the initial version)


samysweb:
Rebased to the current version and ready for review @MatthewDaggitt

I managed to get the tensor example to work. All golden tests are compilable in Isabelle (I've also added disabled golden tests, which check whether Isabelle compiles the output of Vehicle).

docs/exporting.rst
```diff
@@ -138,3 +138,28 @@ directly interface with mathcomp's structure hierarchy. This can lead to issues
...
+Network Evaluation
+##################
+
+Neural networks are represented as fixed parameters of the locale and cannot be evaluated
+within Isabelle.
+
+No integration with verification cache
+######################################
+There is currently no integration with Vehicle's verification cache,
+meaning that it is up to the user to garuntee that the compiled specification
```
MatthewDaggitt: guarantee

vehicle-isabelle/.gitignore
```diff
@@ -0,0 +1 @@
+*~
```
MatthewDaggitt: Rather than adding lots of extra `.gitignore` files scattered throughout, can we add these to the top-level .gitignore file?

vehicle/tests/golden/compile/andGate/test.json
```diff
@@ -23,6 +23,32 @@
...
+      "spec.vcl"
+    ],
+    "produces": [
+      "isabelle/Isabelle.thy"
+    ]
+  },
+  {
+    "name": "IsabelleVerify",
+    "run": "vehicle compile itp -s spec.vcl -t Isabelle -o isabelle/Isabelle.thy && isabelle build -d vehicle-isabelle/ -d isabelle/ IsabelleTest > /dev/null 2>&1 && echo $? || echo 1",
+    "enabled": false,
```
MatthewDaggitt: These should be enabled, but to stop then running when issabelle is not available, you can add the `external` field (see the Rocq example above).

vehicle/tests/golden/compile/andGate/isabelle/ROOT
```diff
@@ -0,0 +1,7 @@
+session IsabelleTest = HOL +
```
MatthewDaggitt, line 1: Are these ROOT files needed to be generated manually for each test?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+-- Isabelle-specific options
+
+data IsabelleOptions = IsabelleOptions
+  { output :: Maybe FilePath,
+    localeName :: Maybe String
+  }
+
+currentPhase :: Doc ()
+currentPhase = "compilation to Isabelle"
+compileProgToIsabelle :: (MonadCompile m) => Prog DecidabilityBuiltin -> IsabelleOptions -> m (Doc a)
```
MatthewDaggitt, line 55: Is there a reason the formatter isn't automatically putting an empty line between `currentPhase` and `compileProgToIsabelle`?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+compileProgToIsabelle (Main ds) options =
+  logCompilerSection2 MinDetail currentPhase $ do
+    logDebug MaxDetail $ prettyExternal (Main ds)
+    -- Combine the printed documents
+    -- let combined = fold gatheredStatements
```
MatthewDaggitt: Can remove this unneeded code?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+            )
+
+    return isabelleProgram
+
+
+-- | Collect dependencies from a 'Code' document by discarding precedence
+--   and folding all dependency annotations.
+collectCodeDependencies :: Code -> Set Dependency
+collectCodeDependencies doc =
+  let stream = layoutPretty defaultLayoutOptions doc
```
MatthewDaggitt, line 95: It's usually cleaner to write lets like this using do notation, e.g.
```
collectCodeDependencies doc = do
  let stream = layoutPretty defaultLayoutOptions doc
  fold (reAnnotateS fst stream)
```

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+   in fold (reAnnotateS fst stream)
+
+-- | Collect dependencies arising from a locale definition. This ensures
+--   that imports required by statements printed in the postamble are
+--   included in the preamble import list.
+collectLocaleDependencies :: [LocaleDef] -> Set Dependency
+collectLocaleDependencies = Set.unions . fmap deps
+  where
+    deps :: LocaleDef -> Set Dependency
+    deps (NetworkDefStatement name ty) =
```
MatthewDaggitt: Usually cleaner to use `\case`, i.e.
```
deps = \case
    NetworkDefStatement name ty ->  collectCodeDependencies name `Set.union` collectCodeDependencies ty
    PropertyDefStatement stmt -> collectCodeDependencies stmt
    TensorTypeDefStmt _ shape body -> collectCodeDependencies shape `Set.union` collectCodeDependencies body
    IndexTypeDefStmt _ maxI body -> collectCodeDependencies maxI `Set.union` collectCodeDependencies body
```
Then you can rename the definition without having to touch so many lines of code.

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+
+data Dependency
+  = RequireImport Library
+  deriving (Eq, Ord)
+
+data LocaleDef
+  = NetworkDefStatement Code Code
+  | PropertyDefStatement Code
+  | TensorTypeDefStmt Identifier Code Code
+  | IndexTypeDefStmt Identifier Code Code
```
MatthewDaggitt, line 136: Can we have empty lines between declarations please?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+gatherLocaleNetworks :: (MonadIsabelleCompile m) => IsabelleOptions -> Decl DecidabilityBuiltin -> m [LocaleDef]
+gatherLocaleNetworks _opts = \case
+  DefAbstract _ n _ t -> do
+    cExpr <- compileExpr False [] t
+    pure [(compilePostulate (compileIdentifier n) cExpr)]
+  _ -> pure []
+
+gatherLocaleStatements :: (MonadIsabelleCompile m) => IsabelleOptions -> [LocaleDef] -> Decl DecidabilityBuiltin -> m [LocaleDef]
+gatherLocaleStatements _opts localeNets = \case
+  -- DefFunction _ n _ (Universe _ _) e -> do
```
MatthewDaggitt: Delete unneeded code?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+  DefAbstract _ _ _ _ -> do
+    return "" -- Done via localeAssm gathering
+  DefFunction _ n anns t e -> do
+    let (binders, body) = foldDeclBinders e
+    if isAnnotatedAsProperty anns
+      then return "" -- Done via localeAssm gathering
+      else compileFunDef localeAssms n t binders body
+        --bindersT bindersV cbody
+  DefRecord _ n _ _t fs -> do
+    -- t' <- compileExpr False localeAssms t
```
MatthewDaggitt: Unneeded code?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+  Meta _ _ -> []
+  Hole _ _ -> []
+
+compileExpr :: (MonadIsabelleCompile m) => Bool -> [LocaleDef] -> Expr DecidabilityBuiltin -> m Code
+compileExpr isOutType localeAssms expr = do
+  logEntry expr
+  result <- case expr of
+    Hole {} -> resolutionError currentPhase "Hole"
+    Meta {} -> resolutionError currentPhase "Meta"
+    Universe _ l -> developerError $ "compilation of universes to Isabelle unsupported: " <> pretty l
```
MatthewDaggitt: Should this call `compileType` instead?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+        let (binders, body) = foldBinders PiBinder binder result
+        compileTypeLevelQuantifier localeAssms Forall (binder :| binders) body
+    Let _ bound binder body -> do
+      cBoundExpr <- compileLetBinder localeAssms (binder, bound)
+      cBody <- addNameToContext binder $ compileExpr False localeAssms body
+      return $ "let" <+> cBoundExpr <+> "in" <+> cBody
+    Lam _ binder body -> compileLam localeAssms binder body
+    Builtin _p b -> compileBuiltin isOutType localeAssms b []
+    App fun args -> compileApp isOutType localeAssms fun args
+    Record _p _i fs -> do
```
MatthewDaggitt: This is quite complicated. Can we split it out into a `compileRecord` function?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+   [LocaleDef] -> [Binder DecidabilityBuiltin] -> m [Code]
+compileTopLevelBindersOuter compileLocaleBinders compileTopLevelBinder localeAssms binders = do
+  let localeResults = compileLocaleBinders localeAssms
+  funBinders <- compileTopLevelBinders compileTopLevelBinder localeAssms binders
+  if (length funBinders == 0)
+    then return []
+    else return $ localeResults ++ funBinders
+
+compileLocaleBindersT :: [LocaleDef] -> [Code]
+compileLocaleBindersT =  (mapMaybe compileLocaleBindersTMapper)
```
MatthewDaggitt: Brackets not needed.

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+  where
+    compileLocaleBindersTMapper (NetworkDefStatement _ t) = Just (parens t)
```
MatthewDaggitt: Can we use `\case` here and below.

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+resolveReturnType localeAssms (_ : bs) (Pi _ binder r) = addNameToContext binder $ resolveReturnType localeAssms bs r
+resolveReturnType localeAssms _ e = compileExpr True localeAssms e
+
+compileRecordField :: (MonadIsabelleCompile m) => [LocaleDef] -> RecordField (Expr DecidabilityBuiltin) -> m Code
+compileRecordField localeAssms (field, fieldValue) = do
+  fieldValue' <- compileExpr False localeAssms fieldValue
+  return $ pretty field <+> "=" <+> (parens fieldValue')
+
+compileTensorTypeDef :: Identifier -> Code -> Code -> Code
+compileTensorTypeDef n shape e = ((vsep :: [Code] -> Code) [
```
MatthewDaggitt: Shouldn't need the type annotations on `vsep`?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+compileNatLiteral i = annotate ([], maxPrecedence) $ "(" <> pretty i <> " :: nat)"
+
+compileTensorLiteral :: (a -> Code) -> Tensor a -> Code
+compileTensorLiteral compileElement t = annotate ([RequireImport VehicleTensor], 200) $ case (shapeOf t, toList t) of
+  ([], [x]) -> parens $ "flextensor_from_vec [] [" <+> compileElement x <+> "]"
+  _ -> foldMapTensor compileElement toTensor t
+  where
+    toTensor :: TensorShape -> [Code] -> Code
+    toTensor _ _ = developerError $
+      "Tensor literal compilation not implemented yet" -- TODO: re-enable when we support tensor literals
```
MatthewDaggitt: This feels like a big flaw as tensor literals are used quite a lot. Is there some way we can fake them?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+idxBasedOp localeAssms op args = case args of
+        [(ExplicitArg _ _ (Lam _ binder _body))] -> case (typeOf binder) of
+          -- (App (Builtin _ IndexType) [maxIdx]) ->
+          (App (Builtin _p (StandardBuiltinType IndexType)) [maxIdx]) -> do
+            idxArg <- (compileExpr False localeAssms (argExpr maxIdx))
+            annotateApp localeAssms [RequireImport VehicleTensor] (op<+>idxArg<>" ") args
+          _ -> developerError $ "foreach/forall/exists tensor operations are currently only supports explicit lambda arguments with indexing type"
+        _ -> developerError $ "foreach/forall/exists tensor operations are currently only supports a single lambda argument"
+
+-- Default precedence for standard operations can be found at https://coq.inria.fr/doc/V8.18.0/refman/language/coq-library.html#notations
```
MatthewDaggitt: Incorrect reference?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,897 @@
...
+      <+> (if null bindersV then mempty else " ") <+> hsep bindersV
+      <+> "="
+      <+> "("
+      <+> cbody
+      <+> ") \"")
+
+idxBasedOp :: (MonadIsabelleCompile m) => [LocaleDef] -> Code -> [Arg DecidabilityBuiltin] -> m Code
+idxBasedOp localeAssms op args = case args of
+        [(ExplicitArg _ _ (Lam _ binder _body))] -> case (typeOf binder) of
+          -- (App (Builtin _ IndexType) [maxIdx]) ->
```
MatthewDaggitt: Remove?

MatthewDaggitt (changes requested):
Apologies for the delay in reviewing. Looks great, a few minor comments, but nothing major!

docs/exporting.rst
samysweb: fixed

vehicle-isabelle/.gitignore
samysweb: fixed

vehicle/tests/golden/compile/andGate/test.json
samysweb: fixed

vehicle/tests/golden/compile/andGate/isabelle/ROOT
samysweb: Alternatively, I could create the ROOT file once somewhere and always copy it to the test directory... Would you prefer that?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: I don't know, but I added a new line now (fixed)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed (for all comments)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed (for all occurrences)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed (hopefully everywhere)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed (for all comments)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed (for all comments)

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: I've unified the error message to compileType now...

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: This is still on my TODO list

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: I tried removing the type annotation, but this produces a type error -- I'm not sure why

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: removed

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: fixed

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: Good thing you are reminding me of this: I meant to come back to this when I get the error when running the golden tests, but this never happened.
`compileTensorLiteral` is called for `NatTensorLiteral`/`BoolTensorLiteral`/`RatTensorLiteral` (`StandardBuiltinConstructor`)
But there is also `ConstTensor` (`StandardBuiltinFunction`), which seems to correctly compile tensors (line 653).
It would maybe/probably make sense to extract this to the compileTensorLiteral functions, but can you give me an example of what I need to write in a vehicle spec file to end up in the `NatTensorLiteral`/`BoolTensorLiteral`/`RatTensorLiteral`  case? All the cases from the golden tests seem to end up in the `ConstTensor` case

MatthewDaggitt:
```haskell
x : Tensor Rat [2]
x = [1.0, 2.0]
```
should end up with a `RatTensorLiteral` for the value of `x

MatthewDaggitt:
@samysweb I'm very sorry, there's going to be a small merge conflict with https://github.com/vehicle-lang/vehicle/pull/1022. In particular:

1.  I've removed the `DefRecord` constructor as it is now merged into `DefFunction`. As you can see in the changes to the Rocq and Agda files in that PR, the information about the original form of the definition is now stored in the `funSort` argument to `DefFunction` which you can then case match on.

2. I've added the ability to parameterise records by types. However, this purely for internal use only at the moment, so only unparameterised records should make it through to the ITP backends. Feel free to add support if you want (/can) but at the moment for the Agda/Rocq backends I'm just throwing an error if the parameter telescope is non-empty.

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: So the original version gave me a type error (I'll attach it below).
I now tried out:
```
x : Tensor Real [2]
x = [1.0, 2.0]

y : Tensor Bool [2]
y = [True, False]

numberOfDogs = 6

type Dog = Index numberOfDogs

z : List Dog
z = [1, 2]
```
This yields the following (which works):
```Isabelle
definition x  :: "  R FlexTensor "
  where " x   = ( (combine_subtensors [(flextensor_from_vec [] [ (((1 :: R) )) ]),(flextensor_from_vec [] [ (((2 :: R) )) ])]) ) "

definition y  :: "  bool FlexTensor "
  where " y   = ( (combine_subtensors [(flextensor_from_vec [] [ true ]),(flextensor_from_vec [] [ false ])]) ) "

definition numberOfDogs  :: "  nat "
  where " numberOfDogs   = ( (6 :: nat) ) "

typedef Dog  =  "{ i :: nat. i < (numberOfDogs) }"
unfolding numberOfDogs_def
  by (simp, rule_tac x = "0" in exI, linarith)

(* Type Coercions *)
declare [[coercion Rep_Dog]]
definition to_Dog :: "FlexIndex \<Rightarrow> Dog"
  where[simp]: "to_Dog a = (
    let i = Rep_FlexIndex a
    in (if i < numberOfDogs then Abs_Dog i else undefined))"
declare [[coercion to_Dog]]

definition z  :: "  Dog list "
  where " z   = ( (Abs_FlexIndex 1) # ((Abs_FlexIndex 2) # []) ) "
```

I.e. we are able to compile tensor literals of this structure, this also works for multi-dimensional tensors...
In my understanding this happens because all the tensor literal calls end up in `ConstTensor` in `compileBuiltin` rather than in `NatTensorLiteral` etc.
Are these cases obsolete or are they triggered in another situation?

For reference here is the error message I get on the `dev` branch for the expression above (just in case that's relevant):
```bash
>cabal run vehicle -- compile itp --target Agda -s test2.vcl --output Test.agda
#Error in file 'test2.vcl' at Line 1, Columns 5-11: unable to work out a valid type for the overloaded expression 'TensorTC'.
#The possible options explored were:
#  1. Type
#    - rejected: unable to find a consistent type for the overloaded expression 'TensorTC'. In particular 'Rat' is not equal to '?23 Rat'.
#  2. Type
#    - rejected: unable to find a consistent type for the overloaded expression 'TensorTC'. In particular 'Rat' is not equal to 'Real'.
#  3. Type
#    - rejected: unable to find a consistent type for the overloaded expression 'TensorTC'. In particular 'Rat' is not equal to 'Bool'.
```

MatthewDaggitt: Ah sorry, I meant (`Real` not `Rat`):
```
x : Tensor Real [2]
x = [1.0, 2.0]
```

MatthewDaggitt: Will look at this in a couple of hours.

MatthewDaggitt: You were absolutely right, `StackTensor` was not being evaluated where it should have been. I have fixed this in https://github.com/vehicle-lang/vehicle/pull/1038

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
```diff
@@ -0,0 +1,900 @@
...
+      _ -> do
+        let (binders, body) = foldPiBinders binder result
+        compileTypeLevelQuantifier localeAssms Forall (binder :| binders) body
+    Let _ bound binder body -> do
+      cBoundExpr <- compileLetBinder localeAssms (binder, bound)
+      cBody <- addNameToContext binder $ compileExpr False localeAssms body
+      return $ "let" <+> cBoundExpr <+> "in" <+> cBody
+    Lam _ binder body -> compileLam localeAssms binder body
+    Builtin _p b -> compileBuiltin isOutType localeAssms b []
+    App fun args -> compileApp isOutType localeAssms fun args
```
samysweb, line 440: @MatthewDaggitt, it seems to me record member accesses/projections are now compiled to function applications? Is there any way for me to distinguish whether the function I am applying is a record member access/record projection? I need to be able to distinguish them for my compilation to work for the `simple-logic` example. Right now `myLogic.true` is seemingly treated like a function application of `true` to `myLogic`

P.S.: Does that mean the `RecordProj` case two lines further down is redundant/unused now?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: done

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
samysweb: Thank you, this allowed me to fix this: I am now generating 1-dimensional tensors and then stack them as needed... I saw you deactivated the ITP tests for `simple-vector` in #1038 -- I've followed suit and deactivated the (failing) Isabelle golden tests for this case. The other occurrences of tensors in the golden tests have been updated and still run with Isabelle compilation

samysweb: so this should be resolved now (besides `simple-vector`)

vehicle/tests/golden/compile/andGate/isabelle/ROOT
samysweb: To put this somewhere else, I would need some place to put an `isabelle` directory with a `ROOT` file which I can then provide in the golden test's `"needs"` section (this would copy `isabelle/ROOT` into the temporary test directory) -- would you have a preferred place for this?

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
MatthewDaggitt: Yes, `simple-vector` needs fixing upstream. Our handling of zero-dimensional tensors is still slightly problematic.

vehicle/src/Vehicle/Backend/Isabelle/Compile.hs
MatthewDaggitt: Ah yes right. This isn't a problem in the Agda or Rocq backends because this is actually how record projections are implemented. 

Given that it only affects the Isabelle backend I don't think we want to add this information to the `Identifier` class. Instead what I suggest is that you track which functions are record projections:

1. change `MonadIsabelleCompile` to include a `MonadState (Set Identifier) m` constraint. 
2. in `compileDecl` when you encounter a projection function, call `modify (Set.insert ident)` to add the identifier to the list of projection identifiers.
3. when you come across an `Identifier` when compiling an expression call `isProjectionFn <- gets (Set.member ident)`

Sorry that code may need a tiny bit of fiddling to get the argument order right, but it should morally work! Even better would be to abstract `isProjectionFn` into it's own method declared underneath `MonadIsabelleCompile` called `isProjectionFunction`.

MatthewDaggitt: Yes, `RecordProj` should only be used inside the projection functions so you don't need to handle it. I should probably write a clean-up pass before the individual ITP compilation to avoid all these cases that we are guaranteed not to need...

MatthewDaggitt:
I've also given you write access to the repository which is long overdue! Apologies.

vehicle/tests/golden/compile/andGate/isabelle/ROOT
MatthewDaggitt: No, that's fine. If they are needed then you can leave them here.

samysweb:
@MatthewDaggitt I implemented the record field access as you suggested and figured out a way to bring Isabelle into the CI (with some inspiration from the CI of another project I stumbled across [1]).
I've also merged with the up to date dev branch. From my side this is ready to merge (pending the succesful completion of the CI)

[1] https://github.com/kappelmann/Isabelle-Set

MatthewDaggitt (approved):

MatthewDaggitt (approved):

PR #1022, opened 2025-12-12
Add parameterised records

MatthewDaggitt:
Next step into moving instances out into the builtins file. Pinging @laurenpudz.


vehicle-syntax/src/Vehicle/Syntax/BNFC/Elaborate/External.hs
```diff
@@ -132,102 +126,181 @@ elabDeclGroup anns = \case
...
+
+  let defToAssign (B.FieldDef nam _tk val) = B.FieldAssign nam val
+  let fields' = B.Record $ fmap defToAssign fields
+
+  let (typ, expr)
+        | null telescope = (tokType 0, fields')
+        | otherwise =
+            ( B.ForallT tokForallT telescope (tokType 0),
+              B.Lam tokLambda telescope tokArrow fields'
+            )
```
MatthewDaggitt, line 285: This is the crucial bit about how records are now formatted @laurenpudz. There is no special `DefRecord` constructor any more. They are all just ordinary function declarations.

This is how a function declaration is constructed from a record definition. See comment below for example.

vehicle/tests/golden/compile/simple-record/spec.vcl
```diff
@@ -0,0 +1,23 @@
...
+  { a : x
+  , b : y
+  }
+
+{-
+Delaborated to
+
+Pair : \forall x y -> Set
+Pair = \x y -> { a : x, b : y }
+-}
```
MatthewDaggitt, line 11: This is what above record definition gets elaborated to internally.

PR #1023, opened 2025-12-12
Replace 'gitrev' with 'gitrev-typed' and fix Vehicle library caching

MatthewDaggitt:
`gitrev` has been deprecated as a project so switching to successor project `gitrev-typed`. This also allows us to control reinstallation of standard library better, by computing the hash of the current changes. This fixes some race conditions that were popping up.


MatthewDaggitt:
Can't use this as it doesn't support GHC 8.X.

PR #1026, opened 2025-12-15
Add TensorLike typeclass

laurenpudz:
Add the `TensorLike` typeclass to `Definitions.vcl`
TensorLike has the toTensorLike and fromTensorLike methods.

### DIFF
--- a/vehicle/lib/Definitions.vcl
+++ b/vehicle/lib/Definitions.vcl
@@ -49,8 +49,8 @@ gtRatTensorReduced xs ys = reduceAnd True (xs >. ys)
 --------------------------------------------------------------------------------
 
 record TensorLike r t dims {{isTensorType : IsTensorType t dims}} where
-  { toTensorLike     : r -> tensor
-  , fromTensorLike   : tensor -> r
+  { toTensorLike     : r -> Tensor t dims
+  , fromTensorLike   : Tensor t dims -> r
   }
 
 --------------------------------------------------------------------------------
--- a/vehicle/lib/Definitions.vcl
+++ b/vehicle/lib/Definitions.vcl
@@ -49,8 +49,8 @@ gtRatTensorReduced xs ys = reduceAnd True (xs >. ys)
 --------------------------------------------------------------------------------
 
 record TensorLike r t dims {{isTensorType : IsTensorType t dims}} where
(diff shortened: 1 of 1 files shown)


vehicle/lib/Definitions.vcl
```diff
@@ -44,6 +44,15 @@ geRatTensorReduced xs ys = reduceAnd True (xs >=. ys)
 gtRatTensorReduced : Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
 gtRatTensorReduced xs ys = reduceAnd True (xs >. ys)
 
+--------------------------------------------------------------------------------
+-- TensorLike
+--------------------------------------------------------------------------------
+
+record TensorLike r t dims {{isTensorType : IsTensorType t dims}} where
+  { toTensorLike     : r -> tensor
```
MatthewDaggitt: This should be `r -> Tensor t dims` and likewise for the function below?

MatthewDaggitt (changes requested):

vehicle/lib/Definitions.vcl
```diff
@@ -44,6 +44,15 @@ geRatTensorReduced xs ys = reduceAnd True (xs >=. ys)
 gtRatTensorReduced : Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
 gtRatTensorReduced xs ys = reduceAnd True (xs >. ys)
 
+--------------------------------------------------------------------------------
+-- TensorLike
+--------------------------------------------------------------------------------
+
+record TensorLike r t dims {{isTensorType : IsTensorType t dims}} where
+  { toTensorLike     : r -> Tensor t dims
```
MatthewDaggitt: I would probably call this `toTensor` and `fromTensor`. The current names suggest that the types should be other way around, i..e your converting to `r`?

MatthewDaggitt (changes requested):

vehicle/lib/Definitions.vcl
laurenpudz: Yep that would make way more sense, thanks for the pickup :) That should all be changed over now!

MatthewDaggitt (approved):

PR #1039, opened 2025-12-17
Add fromTensor conversion function

laurenpudz:
Adds generation of the fromTensor conversion function for records marked with `@tensor`


vehicle/src/Vehicle/Compile/Scope/RecordInstances.hs
```diff
@@ -137,6 +140,7 @@ toRecord t =
   }
 
 -}
```
MatthewDaggitt: Probably can delete this comment now? Or turn it into proper documentation e.g.
```
-- | TEXT GOES HERE
```

vehicle/src/Vehicle/Data/Code/DSL.hs
```diff
@@ -238,6 +241,9 @@ constTensor t x dims = builtinFunction ConstTensor @@@ [t] @@ [x, dims]
 stackTensor :: (BuiltinHasStandardData builtin) => DSLExpr builtin -> DSLExpr builtin -> DSLExpr builtin -> NonEmpty (DSLExpr builtin) -> DSLExpr builtin
 stackTensor t d ds xs = builtinFunction StackTensor @@@ [t, d, ds] @@ xs
 
+atTensor :: (BuiltinHasStandardData builtin) => DSLExpr builtin -> DSLExpr builtin -> DSLExpr builtin -> DSLExpr builtin -> DSLExpr builtin -> DSLExpr builtin
+atTensor t d ds tens idx = builtinFunction AtTensor @@@ [t, d, ds] @@ [tens, idx]
```
MatthewDaggitt: Hmm so we do need to care about irrelevance here. Otherwise we're going to generate something that doesn't necessarily type-check when it interacts with other irrelevant code. So I think this needs to be 
```
builtinFunction AtTensor @@@ [t] .@@@ [t, d] @@ [tens, idx]
```

MatthewDaggitt (changes requested):
Otherwise looks great! 

I guess the final thing to add in this section of the code in the next PR is the declaration of the `@instance`?

MatthewDaggitt (approved):

PR #1053, opened 2026-01-13
Fixed issue #1047

### LINKED ISSUE #1047: Rocq backend import fix
In the new version of mathcomp,
```coq
Require Import mathcomp.ssreflect.ssrbool.
Require Import mathcomp.algebra.ssralg.
Require Import mathcomp.ssreflect.ssrnat.
Require Import mathcomp.ssreflect.eqtype.
Require Import mathcomp.ssreflect.order.
Require Import mathcomp.ssreflect.fintype.
Require Import mathcomp.ssreflect.seq.
Require Import mathcomp.algebra.zmodp.
Require Import mathcomp.reals.reals.
```
Should instead be
```coq
```
As the current no longer compiles

lstrsrmn:
Fixed issue #1047 by splitting mathcomp imports into the correct format and grouping imports into "all_" notation

Updated tests to expect the new import notation for Rocq

### DIFF
--- a/vehicle/src/Vehicle/Backend/Rocq/Compile.hs
+++ b/vehicle/src/Vehicle/Backend/Rocq/Compile.hs
@@ -38,7 +38,7 @@ import Vehicle.Syntax.Tensor
   )
 
 --------------------------------------------------------------------------------
--- Rocq-specific optionsjuice
+-- Rocq-specific options
 
 data RocqOptions = RocqOptions
   { output :: Maybe FilePath,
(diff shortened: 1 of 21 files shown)


vehicle/src/Vehicle/Backend/Rocq/Compile.hs
```diff
@@ -38,7 +38,7 @@ import Vehicle.Syntax.Tensor
   )
 
 --------------------------------------------------------------------------------
--- Rocq-specific options
+-- Rocq-specific optionsjuice
```
MatthewDaggitt: This looks like a typo?

MatthewDaggitt (changes requested):
Otherwise looks great!

Do we need to update some bounds in https://github.com/vehicle-lang/vehicle/blob/dev/vehicle-rocq/vehicle-rocq.opam? If we do, could you also add an entry in the CHANGELOG to make clear that the bounds have changed?

MatthewDaggitt (approved):

PR #1060, opened 2026-02-04
Externalize quantifier typeclasses

laurenpudz:
- Externalizes forall and exists quantifiers. 
- Adds all instances previously defined in the DSL to `definitions.vcl`, as well as an instance for tensorLikes, although the actual quantification to be used with this instance is not yet supported.

**Note:** tests currently fail as `forall` appears as `forallTC` in error messages


MatthewDaggitt:
> Note: tests currently fail as forall appears as forallTC in error messages

I've decided that I don't care about this for the moment. I need to overhaul the general typing error messages and this isn't much of a degradation as the message is already pretty bad...

MatthewDaggitt:
I've also gone through and cleaned up the types for the instances. They have exactly the same meaning as the fully explicit types you were writing before, I've just let the automatic generalisation handle much of it.

PR #1065, opened 2026-02-16
Externalize valid network type typeclasses

laurenpudz:
- Externalized `validNetworkType` typeclass, added `validNetworkIOType` typeclass.
- Changed the `validNetworkType` to be `t1 -> t2`, enforcing that `t1` and `t2` are of `validNetworkIOType`.
- Added a `Tensor Real` instance for `validNetworkIOType` externally, added internal generation of an instance of `validNetworIOType` for each `tensorLike` during scope checking
- Added typecheck tests for a `tensorLike -> tensorLike network` and a `tensorLike -> tensor network`


vehicle/src/Vehicle/Data/Builtin/Standard/Type.hs
```diff
@@ -175,12 +175,12 @@ restrictStandardDeclType ::
...
   env <- getFreeCtx (Proxy @Builtin)
   let tc = case declSort of
-        RestrictedProperty -> ValidPropertyType
-        RestrictedParameter s -> ValidParameterType s
-        RestrictedDataset -> ValidDatasetType
-        RestrictedNetwork -> ValidNetworkType
+        RestrictedProperty -> Builtin p (TypeClass ValidPropertyType)
+        RestrictedParameter s -> Builtin p (TypeClass (ValidParameterType s))
+        RestrictedDataset -> Builtin p (TypeClass ValidDatasetType)
+        RestrictedNetwork -> FreeVar p (standardLibIdent "HasValidNetworkType")
```
MatthewDaggitt: Could you make `standardLibIdent "HasValidNetworkType"` a definition in `Vehicle.Libraries.StandardLibrary`?

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -152,3 +153,17 @@ createTensorLikeInstance p recordIdent fieldElementType fieldDimensions fields =
   let functionIdent = Identifier (modulePath recordIdent) functionName
 
   DefFunction p functionIdent (FunctionDecl 1 (Just (AnnInstance Nothing))) recordType functionBody
+
+createValidNetworkIOInstance ::
+  Provenance ->
+  Identifier ->
+  Decl Builtin
+createValidNetworkIOInstance p recordIdent = do
+  let validNetworkTypeIdent = Identifier standardLibraryDefinitionsModulePath "HasValidNetworkIOType"
```
MatthewDaggitt: Same here.

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -152,3 +153,17 @@ createTensorLikeInstance p recordIdent fieldElementType fieldDimensions fields =
...
+  let recordType = fromDSL mempty $ freeVar validNetworkTypeIdent @@ [freeVar recordIdent]
+  let functionBody = Record p recordType []
+
+  let functionName = Text.pack "_" <> nameOf recordIdent <> "HasValidNetworkIOType"
```
MatthewDaggitt, line 165: Should avoid duplicating the text "HasValidNetworkIOType" here and above?

vehicle/tests/golden/compile/tensorLike-network-mixed/spec.vcl
```diff
@@ -0,0 +1,8 @@
+@tensor
```
MatthewDaggitt, line 1: I think the names of these two tests are mixed up. Shouldn't the other test be called "mixed"? Could we also put the "mixed" in the folder name last so that when you sort them alphabetically they appear next to each other?

MatthewDaggitt (changes requested):
Whoa, super quick. Great that it proved to be so easy 🎉

MatthewDaggitt (approved):

PR #1068, opened 2026-02-19
Test tensorLike quantification

laurenpudz:
- added test for quantifying with tensorLikes
 - did minor correction of tensorLike quantification instance - before we were looking for a `HasQuantifier TensorLike r t dims` but we should actually just be looking for `HasQuantifier r` as this is what will show up when the quantifier is used with the record type
 - there are a couple of tests that started failing in this PR. These were tests that expected a specific error to be thrown, as types were used with quantifiers that had no instances for them (Vector and Nat). Typechecking here still fails, but just doesn't list the instances that have been tried (even though all that should be tried have been tried). Not sure if this is something that may be easier for Matthew to look at than me.

- This work builds upon the PR to externalise allowable network types. Please merge that first before this one.


MatthewDaggitt:
Thanks @laurenpudz, do you think you can merge/rebase on the dev branch so I can get a clearer picture of the changes?

PR #1069, opened 2026-02-20
Issue998

lstrsrmn:
Added support for Rocq's constructive reals and merged with backend presedence fix


lstrsrmn:
Tensor indexing currently gets extracted to be bound tighter than in an original spec
```haskell
type InputVector = Tensor Real [2]

type OutputVector= Tensor Real [1]

@network
pk : InputVector -> OutputVector

@property
example : Bool
example = forall x . (pk x) ! 0 <= 5
```

Extracted file in Rocq
```coq
(* WARNING: This file was generated automatically by Vehicle *)
(* and should not be modified manually! *)
(* Metadata: *)
(*  - Vehicle version: 0.23.0+dev *)
(*  - Rocq version: 9.0.0 *)

Require Import vehicle.tensor.
Open Scope ring_scope.
Open Scope order_scope.

Parameter R : realType.

Definition InputVector : Type := 'nT[R]_(2%N :: nil).

Definition OutputVector : Type := 'nT[R]_(1%N :: nil).

Parameter pk : InputVector -> OutputVector.

Axiom example : forall x, pk x ^^ 0 <= const_t (5 : R).
```

This rocq file will not type check because in rocq, ^^ binds tighter than function application.
Needs to be `(pk x) ^^ 0` instead

PR #1071, opened 2026-02-23
Externalize comparison operators

laurenpudz:
Externalizes comparison operators.
Adds new tokens for all functions needed as values for fields of the `HasQuantifier` typeclass (hence lots of new tokens)

Still has a number of test failures 
- same failures as what are occurring in https://github.com/vehicle-lang/vehicle/pull/1070 (since this branch also includes those changes)
- need to fix one failure related to instances accepting non constant dims


MatthewDaggitt:
Okay so I've got to the bottom of this and again this is a problem with the `TensorLike` construct.

There are three instances:
```
HasQuantifier (Index n)
HasQuantifier (Tensor Real ds)
{{ TensorLike r t dims }} -> {{ HasQuantifier (NonCastingTensor t dims) }} -> HasQuantifier r
```
If it is trying to find an instance for `HasQuantifier Int`, the type-checker can reject the first two easily. However the third one is a possibility as you can unify `r` with `Int`, and so it accepts it. 

It then has to try and prove `TensorLike Int r dims` for some `r` and `dims`. Obviously there are no appropriate instances at all as there are no records so the search fails with an empty list of candidates.

Again, I think we need to internalise and actually construct the `HasQuantifier` instance for the concrete record value of `r` inside the scope checker. This is a useful example of what goes wrong, and would be good to write up an example in your thesis as evidence for having tried both approaches?

MatthewDaggitt:
When https://github.com/vehicle-lang/vehicle/pull/1070 is merged in, can we rebase this one and I will re-review it?

MatthewDaggitt:
Okay long story short, this is a much harder fix then I was anticipating and requires us to do forcing properly in the type-checker. In the meantime, I have just accepted the failing test and we'll merge it in.

MatthewDaggitt:
(the forcing problem is completely unrelated to the project!)

PR #1075, opened 2026-03-03
Fixed issue 1074.

### LINKED ISSUE #1074: Indentation with record
When creating a record, it is natural to not indent the closing `}`, however vehicle cannot type check this. This will typecheck:
```haskell
record t where {
    a : Real,
    b : Real,
    c : Real
    }
```
This will not:
```haskell
record t where {
    a : Real,
    b : Real,
    c : Real
}
```

lstrsrmn:
Fixed #1074, by instructing `}` to break the indentation layout rules. Passes all tests, but this could be too extreme a measure. I couldn't find any instance in the tests of using `{}` for implicit arguments, but I imagine it could cause some oddities around the implicit arguments binders.


MatthewDaggitt:
Hmm so you shouldn't be splitting implicit arguments over multiple lines so should be safe :crossed_fingers: 

Let's see and give it a go! Can you add a test case under the name `issue1074`?

MatthewDaggitt (changes requested):
See comment above.

MatthewDaggitt (approved):

PR #1079, opened 2026-03-05
feat: resolve overwrite & add test for issue1078

### LINKED ISSUE #1078: Monomorphisation issue with polymorphic index.
Running `vehicle typecheck -s path/to/spec` for a specification of this form:
```vehicle
variable = 12

smallIndex : Index 13
smallIndex = variable

bigIndex : Index 42
bigIndex = variable

@property
p : Bool
p = (smallIndex == smallIndex) and (bigIndex == bigIndex)
```
produces [this log](https://github.com/user-attachments/files/25770495/typecheck.log) and fails to typecheck due to an issue with the tracking of applications of an index.

ggustavs:
Change how Monomorphisation.hs tracks declarations in order to not lose track of the same variable being used in different contexts to fix [issue 1078](https://github.com/vehicle-lang/vehicle/issues/1078).

### DIFF
--- a/vehicle/src/Vehicle/Compile/Monomorphisation.hs
+++ b/vehicle/src/Vehicle/Compile/Monomorphisation.hs
@@ -144,7 +144,10 @@ handleUsedDecl applications decl = do
       let numberOfApplications = length monomorphisations
       let allFreeVarsInArgs = Set.unions (freeVarsIn . argExpr <$> concat monomorphisations)
       let createNewName = numberOfApplications > 1 || ident `Set.member` allFreeVarsInArgs
-      traverse (performMonomorphisation (p, ident, anns, typ, body) createNewName) monomorphisations
+      monomorphisationResults <- traverse (performMonomorphisation (p, ident, anns, typ, body) createNewName) monomorphisations
+      let (newDecls, substitutions) = unzip monomorphisationResults
+      tell (Map.singleton ident (typ, HashMap.fromList substitutions))
+      return newDecls
     _ -> do
       logDebug MaxDetail "Not monomorphising as an abstract declaration"
       return [decl]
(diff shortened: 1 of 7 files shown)


vehicle/src/Vehicle/Compile/Monomorphisation.hs
```diff
@@ -278,7 +280,9 @@ replacePreviousApplications ::
   m (Prog builtin)
 replacePreviousApplications prog =
   logCompilerSection2 MaxDetail "applying monomorphisation sites" $ do
-    traverse (traverseFreeVarsM (const id) replaceCandidateApplication) prog
+    -- TODO do this in a single traversal
+    prog' <- traverse (traverseFreeVarsM (const id) replaceCandidateApplication) prog
+    traverse (traverseBuiltinsM replaceDerivedApplication) prog'
```
MatthewDaggitt, line 285: This all looks great, except that I'm not quite sure why we need this extra step? The test passes without it? Do you have a different test case that fails without it?

MatthewDaggitt (commented):
Wow, not quite sure how we failed to pick up on the fact that the map wasn't preserving duplicates before, nice find!

vehicle/src/Vehicle/Compile/Monomorphisation.hs
ggustavs: The backward pass (`collectReferences`) collects monomorphisation candidates from both `FreeVar` nodes and `Builtin(DerivedFunction)` nodes (e.g. `forallIndex`, `existsIndex`). Without this second traversal, when a `DerivedFunction` like `forallIndex` gets monomorphised into multiple copies (because it's used with different type arguments), only the `FreeVar` call sites are updated to point to the new names — the `Builtin(DerivedFunction)` call sites are left referencing the original identifier, which no longer exists.

I've added a test case (`issue1078-derived`) that demonstrates this: a spec using `forall` over `Tensor Real [3]` and `Tensor Real [5]`, giving `forallIndex` two distinct monomorphisations. Without line 285, `CheckLinearity` and `CheckPolarity` fail with `Internal scoping error: declaration 'Definitions.forallIndex' not found in scope`.

MatthewDaggitt: Thanks for the explanation!

Issue #792, opened 2024-03-28
Remove support for GHC 8.10 and 9.0


MatthewDaggitt:
According to Haskell these are both deprecated:

https://haskell-language-server.readthedocs.io/en/stable/support/ghc-version-support.html#current-ghc-version-support-status

wenkokke:
Wait, according to the Haskell Language Server team doesn't mean according to the Haskell Foundation or the GHC team!

MatthewDaggitt:
Completed in #1085

PR #1085, opened 2026-03-11
Remove support for GHC 8.10 and 9.0, add support for 9.10 and 9.12

MatthewDaggitt:
Have various libraries I would like to use that don't support 8.* (e.g. see #1023)

### DIFF
--- a/.github/workflows/build-vehicle.yml
+++ b/.github/workflows/build-vehicle.yml
@@ -27,33 +27,33 @@ jobs:
               extra-args: ""
               extra-args-test-golden: ""
         include:
-          # Build with GHC 9.12.2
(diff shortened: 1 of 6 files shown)


MatthewDaggitt:
Fixes #792

MatthewDaggitt:
@wenkokke do you have any objections to this now? 8.X is really really old.

wenkokke:
> @wenkokke do you have any objections to this now? 8.X is really really old.

Not really. Is there anything we gain by removing support for it?

MatthewDaggitt:
Yes, see my opening comment above. There are libraries I would like to use that have never supported 8.10.

PR #1090, opened 2026-03-13
fix: scalar parameter domain bounds

### LINKED ISSUE #1089: Loss backend does not handle scalar non-network free vars correctly during compilation
When compiling loss for a specification like this:
```vehicle
@network
net : Tensor Real [1] -> Tensor Real [1]

@parameter
epsilon : Real

@property
boundedByEpsilon : Bool
boundedByEpsilon = forall x . -epsilon < x ! 0 < epsilon => -epsilon < net x ! 0 < epsilon
```
the compiler fails with `Error: encountered unexpected  non-network free var`.

However, it does successfully compile a specification like this:
```vehicle
@parameter
r : Real

@network
f : Real -> Real

@property
p : Bool
p = f r > 0
```

ggustavs:
Fixes [issue #1089](https://github.com/vehicle-lang/vehicle/issues/1089) by handling the case where a scalar parameter is used as a domain bound in a specification.


vehicle/src/Vehicle/Backend/Loss/Domain.hs
```diff
@@ -604,8 +607,24 @@ compileLinearExpr dims expr = case toRatTensorValue expr of
   VReduceMulRatTensor {} -> unlinearisable
   VReduceMinRatTensor {} -> unlinearisable
   VReduceMaxRatTensor {} -> unlinearisable
-  VMulRatTensor (TensorOp2Args _ _e1 _e2) -> unlinearisable
-  VDivRatTensor (TensorOp2Args _ _e1 _e2) -> unlinearisable
+  VMulRatTensor (TensorOp2Args _ e1 e2) -> do
+    e1' <- compileLinearExpr dims e1
+    e2' <- compileLinearExpr dims e2
+    if Map.null (coefficients e1') && Map.null (coefficients e2')
```
MatthewDaggitt: Pattern matching on `(isConstant e1', isConstant e2')` would be cleaner here!

vehicle/src/Vehicle/Backend/Loss/Domain.hs
```diff
@@ -604,8 +607,24 @@ compileLinearExpr dims expr = case toRatTensorValue expr of
...
+      then do
+        let TensorValue _ v1 = constantValue e1'
+        let TensorValue _ v2 = constantValue e2'
+        return $ constantExpr $ TensorValue dims (runSilentLogger $ evalMulRatTensor (TensorOp2Args dims v1 v2))
```
MatthewDaggitt: You should be able to do:
```
result <- evalMulRatTensor (TensorOp2Args dims v1 v2)
return $ constantExpr $ TensorValue dims result
```
instead of having to invoke the `runSilentLogger` method.

vehicle/src/Vehicle/Backend/Loss/Domain.hs
```diff
@@ -604,8 +607,24 @@ compileLinearExpr dims expr = case toRatTensorValue expr of
...
+      else unlinearisable
+  VDivRatTensor (TensorOp2Args _ e1 e2) -> do
+    e1' <- compileLinearExpr dims e1
+    e2' <- compileLinearExpr dims e2
+    if Map.null (coefficients e1') && Map.null (coefficients e2')
```
MatthewDaggitt: Same comments here.

MatthewDaggitt (changes requested):
Otherwise looks great!

MatthewDaggitt (approved):

PR #1095, opened 2026-03-18
Typechecker support for records as network IO type

DanielEJones:
- Adds support to the type checker for records as network inputs an outputs
- Compiler will generate `validNetworkIOType` instance for each record
- This instance depends on each field also being an instance of `validNetworkIOType`


vehicle/tests/golden/features/network-multipleInputsOutputs/spec.vcl
```diff
@@ -0,0 +1,7 @@
+record R where
```
MatthewDaggitt, line 7: Could we rename this test folder from `record-network` to `network-multipleInputsOutputs`?

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -32,6 +32,38 @@ instance DesugarableBuiltin Builtin where
...
+  Identifier ->
+  RecordFields Builtin ->
+  Decl Builtin
+createRecordHasValidIOTypeInstance p recordIdent fields =  do
+  -- For each record R we want to create a function that looks like:
+  --
+  --   @instance
+  --   recordRHasValidNetworkIOType :
+  --     {{HasValidNetworkIOType t1}} ->
+  --     {{HasValidNetworkIOType tn}} ->
```
MatthewDaggitt: Could we write this as:
```
  --     {{HasValidNetworkIOType t1}} ->
  ...
  --     {{HasValidNetworkIOType tn}} ->
```

MatthewDaggitt (commented):
Thanks @DanielEJones this looks really good and exactly what I had in mind.

Could we add a couple of tests under the error folder as well. e.g. checking that the following should still fail:
```
record R where
  { a : Real
  ; b : Real -> Real
  }

@network
f : R -> R
```

It also occurs to me that the following should fail, but probably doesn't currently:
```
record R where
  { a : Real
  ; b : Real
  }

record S where
  { c : S
  ; d : Real -> Real
  }

@network
f : R -> R
```
i.e. we only want a single layer of records to be allowable. Maybe you can have a think about how we would implement this and we can discuss it in the meeting tomorrow?

MatthewDaggitt:
It also looks like you may not have setup the pre-commit hooks properly. There should be instructions in the `CONTRIBUTING.md` file?

MatthewDaggitt:
Okay I've fixed the bug with monomorphisation. As well as the tweak to the error messages there's now only one failing test for the `Pair` spec. This is a real failure and arises because `Pair` is a record with generic type parameters. This means that the field types are not in-scope for the generated instance. 
```
(define-fun recordPairHasValidNetworkIOType (pi {{ _ : HasValidNetworkFieldType 𝓲1 }} (pi {{ _ : HasValidNetworkFieldType 𝓲1 }} (HasValidNetworkIOType Pair))) {} )
```

To fix this, you will need to prepend the record telescope to the type of the generated instance.
```
(define-fun recordPairHasValidNetworkIOType (pi {{ a : Type}} {{b : Type }} (pi {{ _ : HasValidNetworkFieldType 𝓲1 }} (pi {{ _ : HasValidNetworkFieldType 𝓲1 }} (HasValidNetworkIOType Pair))) {} )
```

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -33,6 +34,59 @@ instance DesugarableBuiltin Builtin where
...
+  --     {{t1}} ->
+  --     ...
+  --     {{t2}} ->
+  --     {{HasValidNetworkFieldType t1}} ->
+  --     ...
+  --     {{HasValidNetworkFieldType tn}} ->
+  --     HasValidNetworkIOType (R t1 ... tn)
+  --   recordRHasValidNetworkIOType t1 ... tn = {}
+  --
+  -- ... where t1 through tn are the types of R's fields.
```
MatthewDaggitt: telescope not field.

MatthewDaggitt (changes requested):

MatthewDaggitt (approved):

PR #1100, opened 2026-03-26
Updated pygments lexer

lstrsrmn:
Updated the pygments lexer to deal with the newer syntax, and to be closer to the grammar used by vehicle in the hope that it is more maintainable.


wenkokke:
@lstrsrmn the reason why the Pygments lexer was not very well structured is that it’s automatically generated from the grammar using BNFC. (See the comment at the top of the file.) So it’s probably worth regenerating
 the file before we make any changes.

lstrsrmn:
After regenerating the pygments lexer from the bnfc grammar, the new one is definitely better. I think the bnfc grammar will never be able to produce a nice lexer, because of the dependent types, the grammar does not make a distinction between types and terms to my knowledge. With a custom parser, we can interpret what should be visually seen as a term or type where the bnfc grammar cannot. It is my suggestion to extend this PR with some CI that, whenever the bnfc grammar is changed, it somehow prompts a change of the pygments lexer too so no changes slip through the cracks.

PR #1101, opened 2026-03-29
fix: various blocking and infinite looping issues

ggustavs:
Fixes an unblocking issue, infinite loop, and adds index only quantifier support for bounding an index. Adds test to make sure index only quantifier support is not lost


vehicle/src/Vehicle/Compile/Unblock.hs
```diff
@@ -46,22 +45,7 @@ unblockBoolExpr ::
...
-
-  unblockedExpr <- unblockBoolValue actions expr
-
-  newCtx <- getNameContext
-  let unblockedExprDoc = prettyFriendly (WithContext unblockedExpr newCtx)
-  when (layoutAsString exprDoc == layoutAsString unblockedExprDoc) $
-    developerError $
-      "Failed to unblock expression:" <+> exprDoc
-
-  decrCallDepth
```
MatthewDaggitt, line 63: Is there a reason that the logging has been taken out in various places, including here and elsewhere?

vehicle/tests/golden/features/indexOnlyQuantifier/DL2Loss.vcl
```diff
@@ -0,0 +1,3 @@
+@property;
+indexOnlyQuantifier : Tensor Real;
+indexOnlyQuantifier = [ 1.0, 1.0, 1.0, 1.0, 1.0, 1.0 ] / search[x] (\ e -> \ xs -> reduceMul e xs) [6] [ 1.0, 0.0, 0.0, 0.0, 0.0, 0.0 ] [ 1.0, 1.0, 1.0, 1.0, 1.0, 1.0 ] (\ x -> reduceAdd 0.0 ([0.0, 0.0, 1000000.0, 1000000.0, 1000000.0, 0.0] * (max [ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 ] ([ 0.1, 0.1, 0.1, 0.1, 0.1, 0.1 ] - x) + max [ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 ] (x - [ 0.2, 0.2, 0.2, 0.2, 0.2, 0.2 ]))))
```
MatthewDaggitt: Should this file have been committed?

vehicle/src/Vehicle/Backend/Loss/LossCompilation.hs
```diff
@@ -418,9 +419,8 @@ logConversion e action = do
 
   result <- action
 
+  logDebug MaxDetail "post-action"
   decrCallDepth
-  logDebugM MaxDetail $ do
-    outputDoc <- prettyFriendlyInCtx result
-    return $ "exit-loss" <+> ": " <+> outputDoc
+  logDebug MaxDetail "post-decrCallDepth"
```
MatthewDaggitt: I think the old logging was better?

vehicle/src/Vehicle/Backend/Loss/JSON.hs
```diff
@@ -394,12 +381,11 @@ arityError fun arity explicitArgs =
 
 showEntry :: (MonadJSON m) => Value LossBuiltin -> m ()
 showEntry e = do
-  logDebug MaxDetail $ "json-enter:" <+> prettyVerbose e
+  let _ = e
   incrCallDepth
 
 showExit :: (MonadJSON m) => a -> m ()
 showExit _e = do
-  logDebug MaxDetail "json-exit"
```
MatthewDaggitt, line 402: Again, please can we not remove logging?

vehicle/src/Vehicle/Backend/Loss/JSON.hs
```diff
@@ -101,27 +101,16 @@ data JExpr
...
-
-mapRatio :: (Integral b) => (a -> b) -> Ratio a -> Ratio b
-mapRatio f r = do
-  let num = f $ numerator r
-  let denom = f $ denominator r
-  num % denom
+-- NOTE:
+-- Keep JSON rationals unbounded to avoid internal overflows during conversion.
+-- Downstream backends can choose their own finite-precision lowering.
+type Rat = Ratio Integer
```
MatthewDaggitt, line 107: Presumably this change requires a downstream change in the Python parser as well?

vehicle/src/Vehicle/Backend/Loss/Domain.hs
```diff
@@ -270,9 +336,23 @@ findVarBound var VariableInfo {..} (NormalisedRelation rel expr)
   | not (expr `containsVariable` toSliceVar var) = return Nothing
   | otherwise = do
       let (coef, expr') = rearrangeExprToSolveFor (toSliceVar var) expr
-      let boundExpr = tensorValueLinarExprToValue expr'
-      let bounds = convertToTensorBounds parentShape indices rel coef boundExpr
-      return $ Just bounds
+      -- Only extract bounds that are independent of the quantified tensor itself.
+      -- If the bound still depends on any slice from the same parent tensor,
+      -- keep it as a residual constraint in the predicate instead.
```
MatthewDaggitt, line 341: I think this is wrong. We want to keep all bounds, e.g. for a `x : Tensor Real [2]`, then we want to be able to extract bounds from `0 <= x ! 0 <= 1 and 0 <= x ! 1 <= 1`. Won't this stop us from doing so?

vehicle/src/Vehicle/Backend/Loss/Domain.hs
```diff
@@ -67,7 +67,36 @@ checkFinalPartitionUnconstrained ::
 checkFinalPartitionUnconstrained = \case
   (Nothing, Nothing) -> developerError "Found unexpected trivial partition"
   (Nothing, Just value) -> return value
-  (Just {}, _) -> developerError "Constraints still unexpected present after compiling top-level quantifier"
```
MatthewDaggitt, line 70: This shouldn't happen. Constraints are only added to the LHS of a partition if they only involve the quantified variables... If it is happening, then it suggests a bug further down during compilation. 

We already have code for what you've called "closing" the context. It happens automatically in the conversion to loss builtins in `convertBoundVar`.

MatthewDaggitt (commented):
Could we split this out into two PRs? One that changes the representation of rationals (which I think is good to go, modulo changes on the Python side), and one to fix the infinite loop problem (which I think requires more thought)?

vehicle/src/Vehicle/Compile/Unblock.hs
ggustavs: Was fighting debrujin index scoping issues and forgot to add them back, will fix it :)

vehicle/tests/golden/features/indexOnlyQuantifier/DL2Loss.vcl
ggustavs: nope, you are right. For some reason I thought the test failed when I removed it, must have been something else I misattributed to it

vehicle/src/Vehicle/Backend/Loss/LossCompilation.hs
ggustavs: Same as above

vehicle/src/Vehicle/Backend/Loss/JSON.hs
ggustavs: same as above

vehicle/src/Vehicle/Backend/Loss/JSON.hs
ggustavs: will take a look, good point.

vehicle/src/Vehicle/Backend/Loss/Domain.hs
MatthewDaggitt: Oh hang on. This is more complicated than I thought. What concretely is the constraint that is not currently being excluded correctly?

ggustavs: It would fail for a constraint like `x ! 24 + x ! 27 + x ! 30 <= c`, since solving for any of them would involve other slices, which messed up later stages since they assume pure box bounds. Keeping them as residual predicate constraints avoided that mess. Do you have a better solution in mind?

vehicle/src/Vehicle/Backend/Loss/Domain.hs
ggustavs: Kept getting infinite loops for constraints like `x ! 0 + x ! 1 <= c` for the same reason as in the below thread reply

vehicle/src/Vehicle/Backend/Loss/JSON.hs
ggustavs: Already handled by the backend :)

MatthewDaggitt:
Okay I've fixed the unblocking bug in #1103 with a much more minimal fix. The test specification in this PR now passes in #1103.

I cannot reproduce the issue with `x ! 0 + x ! 1 <= c` constraints. Do you have an example specification?

MatthewDaggitt:
@ggustavs are there any more bugs here? Otherwise, I suggest we rescope this PR to just change the floating point representation?

ggustavs:
> @ggustavs are there any more bugs here? Otherwise, I suggest we rescope this PR to just change the floating point representation?

It seems that the recent changes fixed it too? The current issue I get both on the current dev HEAD, this branch, and the latest PyPI release is described [here](https://github.com/vehicle-lang/vehicle/issues/1109)

And yeah, you are right, I will rescope it to just change the floating point representation.

ggustavs:
I will close this PR and open a new one with just the JSON change to keep the commit history cleaner

PR #1106, opened 2026-04-01
Internalise tensorlike instances

laurenpudz:
Removes externally defined instances for tensorLikes, instead generating instances internally, per record. Also removes the generation of an instance for tensorLike for each record, and the definition of the tensorLike typeclass.


MatthewDaggitt (approved):

MatthewDaggitt:
Looks great! Thank you.

PR #1111, opened 2026-04-10
Issue1109

### LINKED ISSUE #1109: Unblock entry network loop
When running `vehicle compile loss -s path/to/`[spec](https://github.com/user-attachments/files/26533070/bug.vcl.txt)` -l DL2Loss`, an `unblock-entry: classifier` loop is produced, as can be seen [here](https://github.com/user-attachments/files/26533252/bug.log.txt)

This behavior is reproducible with both the current PyPI release, and the dev HEAD.

ggustavs:
Fixes #1109 and delegates unblocking recursion decisions to each backend.


MatthewDaggitt (approved):

MatthewDaggitt:
This looks great! Can you add a CHANGELOG entry describing that we fixed a bug in the loss backend then I'm happy to merge this in!

PR #1120, opened 2026-04-18
Add tensorLike datasets

laurenpudz:
- Externalized all dataset-related instances 
- Added the ability to have datasets of tensorLike types - note that this allows datasets with the overall type matching the tensorLike, not tensor datasets with elements of tensorLikes. Eventually we would probably want both allowed? Allowing tensors with tensorLike elements will mean adding instances for `IsTensorType` which would be in a separate PR that I am anticipating on putting up very soon. Also happy to leave this PR if we only want tensor datasets with tesorLike elements
- I think we can remove the instance for the `Index` type to be an allowed tensor dataset element as tensors with `Index` elements aren't allowed? So far it has been left in but I can take it out very easily
- Added simple test
- Had to add default marker to one instance for `ValidDatasetTensorElementType`


vehicle/lib/Definitions.vcl
```diff
@@ -194,6 +194,64 @@ realTensorHasComparison = { leTC = compareRatTensorReducedLe
...
+@instance
+tensorHasValidDatasetListElementType : {{HasValidDatasetTensorElementType t}} -> {{ IsTensorType t dims }} -> HasValidDatasetListElementType (Tensor t dims)
+tensorHasValidDatasetListElementType = {}
+
+@instance
+indexHasValidDatasetListElementType : HasValidDatasetListElementType (Index n)
+indexHasValidDatasetListElementType = {}
+
+@instance
+natHasValidDatasetListElementType : HasValidDatasetListElementType Nat
```
MatthewDaggitt, line 230: This looks as if it should include `real`?

vehicle/src/Vehicle/Data/Builtin/Standard/IndexSolver.hs
```diff
@@ -29,7 +29,7 @@ solveIndexConstraint ::
   m ()
 solveIndexConstraint constraint = do
   normConstraint <- substMetaVariables constraint
-  logDebug MaxDetail $ "Forced:" <+> prettyFriendly normConstraint
+  logDebug MidDetail $ "Forced:" <+> prettyFriendly normConstraint
```
MatthewDaggitt: Could we revert this change?

vehicle/tests/golden/features/tensorLike-dataset/spec.vcl
```diff
@@ -0,0 +1,8 @@
+@tensor
+record Pair where
+  { a : Real
+  , b : Real
+  }
+
+@dataset
+data : Pair
```
MatthewDaggitt: Can we also test that you can create datasets of type `List Pair` and `Vector Pair` as well?

vehicle/tests/golden/features/tensorLike-dataset/spec.vcl
```diff
@@ -0,0 +1,8 @@
+@tensor
```
MatthewDaggitt: Can we call this test `dataset-records`?

MatthewDaggitt (changes requested):
Unfortunately we also need to handle this in the `ExpandResources.Dataset` modules... will discuss!

MatthewDaggitt:
> I think we can remove the instance for the Index type to be an allowed tensor dataset element as tensors with Index elements aren't allowed? So far it has been left in but I can take it out very easily

Yes, that seems fine!

MatthewDaggitt:
> Added the ability to have datasets of tensorLike types - note that this allows datasets with the overall type matching the tensorLike, not tensor datasets with elements of tensorLikes. Eventually we would probably want both allowed? Allowing tensors with tensorLike elements will mean adding instances for IsTensorType which would be in a separate PR that I am anticipating on putting up very soon. Also happy to leave this PR if we only want tensor datasets with tesorLike elements

Ah crucially, I think the primary use case of this is to have these as elements of a `List` or `Vector` type. I don't think we'll have datasets that consist entirely of a single record type.

vehicle/lib/Definitions.vcl
laurenpudz: I think that `real` gets covered by the `real` tensor element case, as it exists as a zero dim real tensor. Tests involving `List Real` pass as is, and I think that the original instances lacked a case for `real` as a list element type? More than happy to add it for completeness though if you would like!!

MatthewDaggitt: Ah yup okay! No need to add

MatthewDaggitt (approved):

PR #1123, opened 2026-04-23
Signal Temporal Logic and closed-loop controllers

ggustavs:
Adds STL temporal operators to the loss backend and the plumbing to specify closed-loop controllers directly in VCL. The computation is efficiently evaluated with [STLCG++](https://uw-ctrl.github.io/stlcg/).

The three operators are `globally[a,b]`, `finally[a,b]`, and `until[a,b]`. They apply to boolean-tensor signals and compose with `forall` / `foreach`. Bounds can be any `Nat` expression that reduces to a literal at compile time, e.g. `globally[0, T - 1]` works when `T : Nat = 10`.

For closed-loop specs `rollout[T] controller dynamics initState` evaluates the controller and the dynamics function together for `T` steps and returns the trajectory as a `Tensor Real [T, stateDim]`. The length of the unrolled sequence corresponds to the initState + `T - 1` transitions. The new `@dynamics` annotation marks the dynamics function, and `transpose` makes it easy to slice along time. 

Added STL as a builtin differentiable logic.
The new `vehicle-stl` submodule under `vehicle-python/` is what handles the computation of the temporal operators. The runtime `Semantics` object is derived from the logic's `pointwiseConjunction` / `pointwiseDisjunction`, which means DL2 and Vehicle can use the temporal operators too without any new fields on the record.

Only the loss backend supports the operators. The verifier and ITP backends reject specs that use them.

**Breaking:** `load_specification` on both backends now returns `(declarations, minimise)` instead of just `declarations`. `minimise` is computed as `true_element < false_element` for each logic.   

Two examples land with the PR: [`temporalSafety`](https://github.com/vehicle-lang/vehicle/tree/feat/temporal-operators/examples/temporalSafety/) for an open-loop ACC spec, and [`closedLoopSafety`](https://github.com/vehicle-lang/vehicle/tree/feat/temporal-operators/examples/closedLoopSafety/) for a double-integrator reach-avoid problem. Reference docs at [`docs/language/temporal-operators.rst`](https://github.com/vehicle-lang/vehicle/tree/feat/temporal-operators/docs/language/temporal-operators.rst).

### Running it locally
By default, git does not initialise submodules, so, to use the Python bindings locally, it is necessary to initialise the submodule first:
 
```bash
git submodule update --init --recursive
```


ggustavs:
@MatthewDaggitt reminder to check out if there is a way to move `convertBoolTensor` out of [`Domain.hs`](https://github.com/vehicle-lang/vehicle/blob/4e61e3723d3b1835838f7275864e55c63d8dc0a7/vehicle/src/Vehicle/Backend/Loss/Domain.hs#L54) and into LossCompilation.hs as it should be

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+   :local:
+
+Signal Temporal Logic operators
+-------------------------------
+
+Vehicle provides three *Signal Temporal Logic* (STL) operators for
+expressing properties over discrete-time boolean signals. Each takes
+a bounded interval ``[a, b]`` of step indices and a ``Tensor Bool``
+signal, and returns a ``Tensor Bool`` signal of the same shape.
+The outer dimension of the signal is the time dimension.
```
MatthewDaggitt: Can we add a table for the three operators and their types, similarly to how we do for e.g. boolean operations?

docs/language/index.rst
```diff
@@ -16,6 +16,7 @@ of neural networks.
   tensors
   records
   quantifiers
+  temporal-operators
```
MatthewDaggitt: The pages are designed to be read sequentially, and this page requires knowledge of networks, etc., so I feel that this should come near the end.

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+produces a ``Tensor Real [10, 2]`` trajectory. Take its first column
+to get the sequence of positions, and build an element-wise boolean
+signal recording whether the position at each step is within bounds:
+
+.. code-block:: agda
+
+   @network
+   controller : Tensor Real [2] -> Tensor Real [1]
+
+   @dynamics
```
MatthewDaggitt, line 103: Need to document what the @dynamics flag does, and what types of functions it accepts?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+signal, and returns a ``Tensor Bool`` signal of the same shape.
+The outer dimension of the signal is the time dimension.
+
+The signal is typically derived from a state trajectory. Most STL
+specifications describe *closed-loop* behaviour: a controller and a
+dynamics model are rolled out together to produce the trajectory one
+step at a time. Temporal operators also apply unchanged to
+*open-loop* specifications, where a network predicts the full
+trajectory in one forward pass (see
+`Open-loop specifications`_ below).
```
MatthewDaggitt: How is this related to the rollout operator? What is it's type?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+   trajectory = rollout[10] controller dynamics initState
+
+   positions : Tensor Real [10]
+   positions = (transpose trajectory) ! 0
+
+   belowLimit : Tensor Bool [10]
+   belowLimit = const 0 [10] .< positions and positions .< const 10 [10]
+
+
+Notice the `.<` operator, which is a pointwise comparison that produces a boolean tensor.
```
MatthewDaggitt: Feels like we should fix this... and remember to come back to change the documentation.  Can you open an issue or a PR for it?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+that step is between 0 and 10. Trying to write it like
+``const 0 [10] .< positions .< const 10 [10]`` would not work,
+because one comparison would be interpreted as a boolean tensor,
+and the other comparison would fail with a type error.
+
+
+The ``globally`` operator
+-------------------------
+
+``globally[a, b] signal`` holds at step ``t`` when the signal
```
MatthewDaggitt: Where does `t` appear? It's not bound anywhere?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+that a goal is eventually reached:
+
+.. code-block:: agda
+
+   inGoal : Tensor Bool [10]
+   inGoal = foreach i . goalLo <= positions ! i <= goalHi
+
+   @property
+   reachGoal : Bool
+   reachGoal = (finally[0,9] inGoal) ! 0
```
MatthewDaggitt: Would it make sense to reuse `belowLimit` again rather than defining a new predicate?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+true at some step ``j`` within ``[t + a, t + b]`` and ``phi`` holds
+at every step strictly before ``j``. Use it to combine a progress
+requirement with a safety invariant that must hold until the goal
+is reached:
+
+.. code-block:: agda
+
+   @property
+   safeUntilGoalReached : Bool
+   safeUntilGoalReached = (until[0,9] belowLimit inGoal) ! 0
```
MatthewDaggitt: `inGoal` is not defined anywhere?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+The per-step predicate can itself use a first-order quantifier over
+the state dimensions. Build the boolean signal with ``foreach`` over
+the time dimension and ``forall`` over an ``Index`` type to require
+that every component of the state stays within bounds until the
+position enters a goal band:
+
+.. code-block:: agda
+
+   stateInBounds : Tensor Real [2] -> Bool
+   stateInBounds s = forall k . stateLoBounds ! k <= s ! k <= stateHiBounds ! k
```
MatthewDaggitt, line 192: `StateLoBounds` and `StateHiBounds` not defined?

Should be able to write this as `stateLoBounds <= s <= stateHiBounds`?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+holds at every step in the window ``[t + a, t + b]``. Use it to
+specify that a safety property must hold throughout an interval:
+
+.. code-block:: agda
+
+   @property
+   stayBounded : Bool
+   stayBounded = (globally[0,9] belowLimit) ! 0
+
+The outer ``! 0`` extracts the result at the first time step; because
```
MatthewDaggitt, line 137: This `! 0` pattern appears everywhere in every spec. Does this suggest we're missing an abstraction somewhere? Should the operators automatically return the first component?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+------------------------
+
+Not every specification needs a dynamics model. An *open-loop*
+controller predicts the full trajectory in one forward pass, and the
+temporal operators apply to its output directly:
+
+.. code-block:: agda
+
+   @network
+   controller : Tensor Real [2] -> Tensor Real [10]
```
MatthewDaggitt: Hmm I think we should rename `controller`? Instead should be called `trajectoryPrediction` or something similar? It contains both the controller and the dynamics right?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   vTrace : Tensor Real [10]
+   vTrace = controller initState
+
+   belowLimit : Tensor Bool [10]
+   belowLimit = const 0 [10] .<= vTrace and vTrace .<= const 10 [10]
```
MatthewDaggitt: Maybe more readable as `foreach i . 0 <= vTrace ! i <= 10`?

docs/language/temporal-operators.rst
```diff
@@ -0,0 +1,173 @@
...
+   alwaysBelowLimit : Bool
+   alwaysBelowLimit = (globally[0,9] belowLimit) ! 0
+
+The operators themselves, and all the patterns above, work the same
+way in both styles. The choice is about how the trajectory is
+produced, not about which properties can be expressed over it.
+
+
+Support
+------------------------------
```
MatthewDaggitt: Title underlining too long?

Can we also put this as an alert box at the very top of the page? There's special RST syntax for warning/alert I think?

ChangeLog.md
```diff
@@ -4,6 +4,14 @@
 
 ### Loss backend
 
+* **Breaking (Python API):** `load_specification` now returns a tuple
+  `(declarations, minimise)`. The `minimise` flag comes from the compiled
+  logic's direction — `True` for loss-style logics (`DL2`, `Vehicle`)
+  whose output should be minimised directly, and `False` for
+  robustness-style logics (`STL`) whose output should be negated before
+  being minimised. Training loops that branch on this flag become
+  portable across differentiable logics.
```
MatthewDaggitt: This feels ugly. Is there a reason we can't do the flipping internally in Vehicle rather than exposing it to the user?

ChangeLog.md
```diff
@@ -4,6 +4,14 @@
...
+
```
MatthewDaggitt, line 62: Should presumably add an entry for the new temporal logic support as well?

vehicle/src/Vehicle/List.hs
```diff
@@ -95,6 +95,7 @@ searchDecl decl = do
   case decl of
     DefAbstract _ _ sort t -> case sort of
       NetworkDef -> tell [Network $ NetworkSummary (sharedData t)]
+      DynamicsDef -> tell [Network $ NetworkSummary (sharedData t)]
```
MatthewDaggitt: This seems wrong... Can we create it's own constructor?

vehicle/lib/Definitions.vcl
```diff
@@ -78,7 +78,11 @@ record HasSub t1 t2 t3 where
   { subTC : t1 -> t2 -> t3
   }
 
-@instance
+@instance(default=0)
+natHasSub : HasSub Nat Nat Nat
+natHasSub = { subTC = subNat }
```
MatthewDaggitt: Need to add these new definitions to the documentation for `Nat` and add tests. In particular the `div` one...

vehicle-syntax/src/Vehicle/Syntax/External.cf
```diff
@@ -244,6 +252,11 @@ If.      Expr4  ::= TokIf Expr5 TokThen Expr5 TokElse Expr5;
 Impl.    Expr5  ::= Expr6 TokImpl Expr5;
 And.     Expr7  ::= Expr8 TokAnd Expr7;
 Or.      Expr7  ::= Expr8 TokOr  Expr7;
+Globally. Expr7 ::= TokGlobally TokSeqOpen Expr "," Expr TokSeqClose Expr8;
+Finally.  Expr7 ::= TokFinally  TokSeqOpen Expr "," Expr TokSeqClose Expr8;
+Until.    Expr7 ::= TokUntil    TokSeqOpen Expr "," Expr TokSeqClose Expr14 Expr14;
+Rollout.  Expr7 ::= TokRollout  TokSeqOpen Expr TokSeqClose Expr14 Expr14 Expr14;
+Transpose. Expr14 ::= TokTranspose Expr15;
```
MatthewDaggitt, line 260: Need to document and add tests for the transpose operator?

vehicle/src/Vehicle/Data/Builtin/Standard/Normalise.hs
```diff
@@ -85,6 +89,8 @@ instance NormalisableBuiltin Builtin where
       ForeachTensor -> NonSimple evalForeachTensor
       ForeachVector -> NonSimple evalForeachVector
       Iterate -> NonSimple evalIterate
+      ReverseDims -> Simple evalReverseDims
```
MatthewDaggitt: I don't think you need `ReverseDims` as a native operator. Should be possible to implement as a standard library function `reverse xs = foldr (\x acc -> acc ++ [x]) [] xs`

vehicle/src/Vehicle/Data/Builtin/Standard/Core.hs
```diff
@@ -334,6 +344,12 @@ instance BuiltinHasStandardData Builtin where
 instance BuiltinHasIterate Builtin where
   accessIterateBuiltin = functionAccessor Iterate
 
+instance BuiltinHasRollout Builtin where
+  accessRolloutBuiltin = functionAccessor Rollout
+
+instance BuiltinHasTranspose Builtin where
```
MatthewDaggitt: This should go in the `BuiltinHasTensors` typeclass.

vehicle/src/Vehicle/Compile/Sugar/Resugar/Internal.hs
```diff
@@ -53,6 +53,7 @@ instance Delaborate V.DefAbstractSort (B.NameToken -> B.Expr -> B.Decl) where
   delabM sort = return $ case sort of
     V.BuiltinDef {} -> B.DeclPost
     V.NetworkDef -> B.DeclNetw
+    V.DynamicsDef -> B.DeclNetw
```
MatthewDaggitt: The grammar here should be updated, otherwise `prettyVerbose` is going to print the wrong thing.

docs/training.rst
```diff
@@ -6,21 +6,23 @@ Vehicle compiles each ``@property`` in your specification into a callable loss f
...
-Feed a ``.vcl`` file to the backend-specific ``load_specification`` helper. Each property becomes a callable entry whose signature mirrors the ``@network`` declarations referenced inside that property:
+Feed a ``.vcl`` file to the backend-specific ``load_specification`` helper. It returns a pair ``(declarations, minimise)``: a dictionary of compiled callables and a flag describing the sign convention of the selected logic. Each callable entry has a signature that mirrors the ``@network`` declarations referenced inside the corresponding property:
 
 .. code-block:: python
 
    import vehicle_lang as vcl
    from vehicle_lang.loss import pytorch as loss_pt
 
-   declarations = loss_pt.load_specification(
+   declarations, minimise = loss_pt.load_specification(
```
MatthewDaggitt: It feels it would be nicer to do the sign flipping of the logic internally in `load_specification`. Is there a reason we can't do this?

examples/temporalSafety/README.md
```diff
@@ -0,0 +1,71 @@
+# Adaptive Cruise Control — temporal safety example
+
+An ego vehicle follows a lead vehicle.  The controller receives the initial
```
MatthewDaggitt, line 3: What is an "ego vehicle"? Is there a way we explain this in plainer language?

examples/temporalSafety/README.md
```diff
@@ -0,0 +1,71 @@
...
+Learned velocity trace (m/s)  [target band: 28.5 – 31.5]
+  step  0:  xx.xx m/s  ...
+  ...
+  step  9:  ~30.x m/s  ###############
+```
+
+## Logic-portable sign handling
+
+`load_specification` returns `(declarations, minimise)`.  The flag lets a
+single training script run unchanged under any differentiable logic:
```
MatthewDaggitt: This `minimise` flag is infecting almost every file...

examples/temporalSafety/temporalSafety.vcl
```diff
@@ -0,0 +1,78 @@
...
+-- Temporal properties
+
+-- Safety: speed must stay within physical limits [0, vMax] at every step,
+-- and no consecutive step change may exceed maxAccel (law of physics).
+-- VCL's type system does not support i+1 index arithmetic in foreach, so the
+-- nine step-to-step smoothness checks are enumerated explicitly.
+@property
+alwaysBelowLimit : Bool
+alwaysBelowLimit = ((globally[0,9] belowLimit) ! 0)
+    -- and (forall i . -maxAccel <= (vTrace ! i) - (vTrace ! (i-1)) <= maxAccel)
```
MatthewDaggitt: Remove the commented out code?

examples/temporalSafety/train.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+"""Adaptive Cruise Control: training a controller to satisfy temporal properties.
+
+Scenario
+--------
+An ego vehicle follows a lead vehicle.  The controller receives the initial
```
MatthewDaggitt: Again "ego"

vehicle/src/Vehicle/Backend/ITP/Agda.hs
```diff
@@ -568,7 +568,9 @@ compileBuiltinFunction f args = case f of
   Not -> annotateApp [DataBool] Nothing "not" args
   Implies -> annotateInfixApp [VehicleUtils] 4 Nothing "_⇒_" args
   Add AddNat -> annotateInfixApp [DataNat] 6 (Just natQualifier) "_⊕_" args
+  Sub SubNat -> annotateInfixApp [DataNat] 6 (Just natQualifier) "_∸_" args
   Mul MulNat -> annotateInfixApp [DataNat] 7 (Just natQualifier) "_*_" args
+  Div DivNat -> annotateInfixApp [DataNat] 7 (Just natQualifier) "_/_" args
```
MatthewDaggitt: Hmm this actually isn't guaranteed to work as in Agda we need a proof that the dividend is non-zero which Vehicle can't currently provide. Which spec do we need division of naturals for?

vehicle/src/Vehicle/Backend/Loss/JSON.hs
```diff
@@ -51,10 +55,12 @@ convertFromJSONProg = fromJProg
 -- The AST exported to JSON
 --------------------------------------------------------------------------------
 
-newtype JProg
-  = Main [JDecl]
+data JProg = Main [JDecl] JLogicMetadata
   deriving (Generic)
 
+data JLogicMetadata = JLogicMetadata JExpr JExpr JExpr JExpr L.LogicDirection
```
MatthewDaggitt: This should probably be a record with named fields?

MatthewDaggitt (commented):
Added a bunch of feedback. We should definitely have that meeting with Wen though. 

The other point is what happens when you use some of these operators in the other backends? Both the temporal operators and the non-temporal operators. For example, I didn't see any indication of handling transpose in the query backend?

ChangeLog.md
ggustavs: We can, it just would shift Vehicle from translating the property with a DL and returning it as an executable function, to it translating and potentially negating the property to ensure the returned function must be minimised, no matter the direction of the DL.

This could be fine for most applications, but annoying for some others. For example, if you want to use Vehicle to write a property describing good behaviour and translate it with a DL, for which [[True]] > [[False]], then it would be quite unintuitive for the returned function to be the opposite of what they expect. If they want to have their intended semantics, they have to accept double-negation.

vehicle/src/Vehicle/Backend/ITP/Agda.hs
ggustavs: It's mostly for convenience when defining intervals of temporal operators. For example, if you define T to be the length of the entire signal, you might want to write something like `finally[0,T/2] phi`

The issue of DivNat affecting multiple backends could be avoided by adding a temporal/ time type, which could be used for defining the temporal operator interval bounds instead.

docs/language/arithmetic.rst
```diff
@@ -14,38 +14,96 @@ The type of natural numbers is written as ``Nat``.
...
+   * - Max
+     - :code:`max`
+     - :code:`Nat -> Nat -> Bool`
+     - :code:`max x y`
 
 Note that inequalities can be chained, so that ``x < y <= z`` will be
 expanded to ``x < y and y <= z``.
 
+Time
+----
```
MatthewDaggitt, line 65: Hmm I think this probably better lives in the `Temporal operators` section because it's not like anyone can use this without importing STL?

docs/language/networks.rst
```diff
@@ -44,6 +42,27 @@ There are various libraries to convert trained models to the ONNX format:
 
 If you would be interested in other formats being supported, please get in touch.
 
+The ``@dynamics`` annotation
+----------------------------
```
MatthewDaggitt, line 46: Again probably best to go in the `temporalOperators` section?

docs/language/tensors.rst
```diff
@@ -66,170 +66,94 @@ Operations
...
+``[T, stateDim]`` into ``[stateDim, T]`` so that an individual
+state component can be sliced as ``Tensor Real [T]``.
+
+``transpose`` is normalised away at compile time when the result is
+indexed: ``(transpose t) ! i ! j`` reduces to ``t ! j ! i``. Literal
+tensors (e.g. ``transpose [[1.0, 2.0], [3.0, 4.0]]``) are folded at
+compile time as well.
+
+Backend support (non-loss)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
MatthewDaggitt, line 147: We now have nice badges to illustrate support. See the rest of the file.

docs/language/tensors.rst
```diff
@@ -66,170 +66,94 @@ Operations
...
+
+``transpose t`` reverses the dimension order of a tensor:
+``Tensor A [d_1, d_2, ..., d_n]`` becomes ``Tensor A [d_n, ..., d_2, d_1]``.
+For a 2-tensor this is the standard matrix transpose. It is used in
+the temporal-operators running example
+(see :doc:`temporal-operators`) to project a trajectory of shape
+``[T, stateDim]`` into ``[stateDim, T]`` so that an individual
+state component can be sliced as ``Tensor Real [T]``.
+
+``transpose`` is normalised away at compile time when the result is
```
MatthewDaggitt, line 141: I don't think this paragraph is neither true (in some scenarios it may not be eliminated) nor relevant to the user.

docs/language/tensors.rst
```diff
@@ -66,170 +66,94 @@ Operations
...
+
+Transpose
+---------
+
+``transpose t`` reverses the dimension order of a tensor:
+``Tensor A [d_1, d_2, ..., d_n]`` becomes ``Tensor A [d_n, ..., d_2, d_1]``.
+For a 2-tensor this is the standard matrix transpose. It is used in
+the temporal-operators running example
+(see :doc:`temporal-operators`) to project a trajectory of shape
+``[T, stateDim]`` into ``[stateDim, T]`` so that an individual
```
MatthewDaggitt, line 138: Again, maybe put this last sentence in the temporal documentation but not here. Please can you fold the operator description into the table.

docs/language/tensors.rst
```diff
@@ -66,170 +66,94 @@ Operations
...
-     - | ``Tensor A ds → A``
-       | (if ``A`` supports ``max``)
-     - |backendall_full|
+     - Pointwise subtract the values in the first tensor from the values
+       in the second. Only valid if subtraction is defined for the type of
+       elements ``A``.
 
+Pointwise comparisons chain like the elementwise forms: ``a <=. b <=. c``
+is sugar for ``a <=. b and b <=. c`` and produces a ``Tensor Bool ds``.
+Use this for interval bounds, e.g. ``stateLoBounds <=. s <=. stateHiBounds``
```
MatthewDaggitt, line 126: Please put this sentence in the `temporal-operators` section rather than here?

MatthewDaggitt (commented):
Some comments on the documentation. Basically I think we should unify everything to do with temporal operators under the temporal operators page!

PR #1129, opened 2026-05-06
Updating vehicle-rocq backend to tensor PR + caching

hoheinzollern:
- reference tensor PR : math-comp/math-comp#1535 (currently compiles agains that branch)
- `tensor.v` dropped, updated codegen to match
- new tensor library has ergonomic notations for tensors, used in the new codegen
- tested against windController and `SafetyProof.v` is updated to the new tensor API
- updated golden tests to match
- added caching support for the Rocq backend


vehicle/tests/golden/specifications/increasing/Rocq.v.golden
```diff
@@ -6,13 +6,14 @@
...
-Open Scope ring_scope.
-Open Scope order_scope.
+Local Open Scope ring_scope.
+Local Open Scope order_scope.
+Local Open Scope tensor_scope.
 
 Parameter R : realType.
 
-Parameter f : 'nT[R]_(nil) -> 'nT[R]_(nil).
+Parameter f : 'nT[R]_([tuple]) -> 'nT[R]_([tuple]).
```
MatthewDaggitt: Just want to check that `tuple` really is the syntax for the empty tuple?

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
 compileNatLiteral :: Int -> Code
 compileNatLiteral i = annotateConstant [MathcompImport Boot] $ pretty i <> "%N"
 
+-- | Compile a tensor type using mathcomp's purely-contravariant shorthand
+-- 'nT[R]_[n1, .., nk] for non-empty dimension lists, or 'nT[R]_([tuple])
+-- for the empty list.
+compileTensorType :: (MonadRocqCompile m) => [Arg DecidabilityBuiltin] -> m Code
+compileTensorType args = case filter isExplicit args of
```
MatthewDaggitt: Can we use:
```
case getExpr accessSpine args of
  Just TensorTypeArgs{..} -> _
  _ -> error ...
```
here instead?

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
...
+-- | Walk a literal Cons/Nil dimension list expression and return the
+-- list of compiled element codes. Errors if the expression isn't a
+-- literal Cons/Nil chain (vehicle's elaborator always reduces tensor
+-- shapes to literal Nat-Cons-Nil chains before codegen).
+compileDimList :: (MonadRocqCompile m) => Expr DecidabilityBuiltin -> m [Code]
+compileDimList = go []
+  where
+    go acc expr = case expr of
+      -- Nil case: end of list
+      Builtin _ (StandardBuiltinConstructor Nil) ->
```
MatthewDaggitt: You should be able to use the pattern synonyms `IDimNil` and `IDimCons` here?

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
...
+      App (Builtin _ (StandardBuiltinConstructor Nil)) _ ->
+        return (reverse acc)
+      -- Cons case: extract head and recurse on tail
+      App (Builtin _ (StandardBuiltinConstructor Cons)) consArgs ->
+        case filter isExplicit (NonEmpty.toList consArgs) of
+          [headArg, tailArg] -> do
+            n <- compileDimElem (argExpr headArg)
+            go (n : acc) (argExpr tailArg)
+          _ -> developerError "compileDimList: malformed Cons in dimension list"
+      _ -> developerError "compileDimList: dimension list is not a literal Cons/Nil chain"
```
MatthewDaggitt: Spaces here? Looks like you haven't installed the pre-commits hooks? See CONTRIBUTING.md.

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
...
+    compileDimElem (Builtin _ (StandardBuiltinConstructor (NatLiteral n))) =
```
MatthewDaggitt: Again should be able to use `INatLiteral n` pattern synonym, and can we have a type for this function?

Actually why do we need this case at all?? Shouldn't `compileExpr` fall through to it?

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
...
+  _ -> developerError "compileTensorType: expected exactly two explicit arguments (element type and dimension list)"
+
+-- | Walk a literal Cons/Nil dimension list expression and return the
+-- list of compiled element codes. Errors if the expression isn't a
+-- literal Cons/Nil chain (vehicle's elaborator always reduces tensor
+-- shapes to literal Nat-Cons-Nil chains before codegen).
+compileDimList :: (MonadRocqCompile m) => Expr DecidabilityBuiltin -> m [Code]
+compileDimList = go []
+  where
+    go acc expr = case expr of
```
MatthewDaggitt: Can we have a type for this function?

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
```diff
@@ -670,8 +693,56 @@ compileIndexLiteral i =
...
+              Open RingScope
+            ],
+          Nothing
+        )
+        body
+  _ -> developerError "compileTensorType: expected exactly two explicit arguments (element type and dimension list)"
+
+-- | Walk a literal Cons/Nil dimension list expression and return the
+-- list of compiled element codes. Errors if the expression isn't a
+-- literal Cons/Nil chain (vehicle's elaborator always reduces tensor
```
MatthewDaggitt: This statement isn't true at all... You can definitely have variables as tensor parameters. e.g.
```
n = 2

Tensor Real [2,n]
```
Does the Rocq tensor library support this or do they have to be literals?

examples/windController/Makefile
```diff
@@ -0,0 +1,25 @@
+# Build the Rocq proof for the windController example.
+#
+# Run from this directory so that the relative paths inside
+# `verificationResult/.vcl-cache-index` (windController.vcl, controller.onnx)
+# resolve correctly when the `vehicle_validate` tactic invokes
+# `vehicle validate --cache=verificationResult` at proof-checking time.
```
MatthewDaggitt: Hmm we should really switch to using an absolute path for the pointer to the cache so that we don't run into this problem...

MatthewDaggitt (changes requested):
Looks generally good, modulo a few comments. Could you also update the "Limitations" section of the ITP file?

MatthewDaggitt:
Oh and can you add a CHANGELOG entry as well?

vehicle/tests/golden/specifications/increasing/Rocq.v.golden
hoheinzollern: Actually, I think we are missing a notation for tensors with empty dimensions, I'm proposing `'T[R]_[]` on the mathcomp end. Will update both ends now. Empty lists, empty tuples and functions from ordinals of cardinality zero (`T ^ 0`) are all isomorphic and have automatic coercions.

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
hoheinzollern: Will do now.

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
hoheinzollern: You're right that `compileExpr` would handle it. `2%N` works inside `'nT[R]_[...]` since this notation automatically enters `nat_scope` via `%:posnat`. Both 'nT[R]_[2] and 'nT[R]_[2%N] typecheck.

I would keep this for readability, as it outputs e.g. `2` instead of `2%N`. We fallback to `compileExpr` for non-literal dimensions. I'll add the type signature.

Let me know you'd rather have one fewer case at the cost of making the output heavier with `%N`.

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
hoheinzollern: I think this comment was stale from me attempting different notations (synced between the two sides). The Rocq tensor library supports literals, but keep in mind that we need to be able to infer that the dimensions are positive (so there's a distinction between `nat` and `{posnum nat}`, the latter is a sigma type carrying a proof of positivity). I have not tested this in cases where we have a tensor size that is a variable, do we have such examples?
`interval_inference` should automate this process, so you could translate `Tensor Real [2, n]` where `n` is natural and infer its positivity, but it might fail when the expression is complex. Should we accept that / add tests?

examples/windController/Makefile
hoheinzollern: Ok, I'll integrate that.

hoheinzollern:
> Oh and can you add a CHANGELOG entry as well?

Definitely, I was waiting for the review :)

hoheinzollern:
Is this comment still relevant:
```
Poor tensor integration with Mathcomp
#####################################

Currently, tensors are implemented using nested mathcomp tuple types and does not
directly interface with mathcomp's structure hierarchy. This can lead to issues
when considering properties with tensor arithmetic. Users are encouraged to, when
possible, express tensor properties using universal quantification over indices.
This generally leads to neater proofs.
```
@MatthewDaggitt can you point me to where this can be an issue?

vehicle/tests/golden/specifications/increasing/Rocq.v.golden
MatthewDaggitt: Sounds good to me!

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
MatthewDaggitt: Nope, I'm all for prettifying the output. The extra line seems worth it.

examples/windController/README.md
```diff
@@ -43,9 +45,51 @@ vehicle export \
...
+relocatable, pass absolute paths for the specification and network when
+running `vehicle verify` (otherwise those resource paths are stored
+relative to the verify cwd and `vehicle validate` will look for them
+there).
+
+To build the Rocq proof:
+
+```bash
+cd examples/windController/rocqProof
+make
```
MatthewDaggitt, line 90: I don't think this make file exists any longer?

MatthewDaggitt (commented):
Otherwise looks great to me!

MatthewDaggitt:
> @MatthewDaggitt can you point me to where this can be an issue?

This comment was written by Josh before the complete rewrite, so if you are not aware of any issues, then yes I think we can treat it as stale and remove it.

MatthewDaggitt:
Also have just invited you to give you write permissions for the repo @hoheinzollern

examples/windController/README.md
hoheinzollern: (the makefile is now autogenerated by rocq makefile)

lstrsrmn:
The extraction does not require import utils when seeing inequalities, although it relies on it to work. Example vcl file:

```haskell
@network
f : Tensor Real [1] -> Tensor Real [1]

@property
test : Bool
test = forall x . 0 < (x ! 0) => 0 < ((f x) ! 0)
```
Generated file:
```coq
(* WARNING: This file was generated automatically by Vehicle *)
(* and should not be modified manually! *)
(* Metadata: *)
(*  - Vehicle version: 0.24.1 *)
(*  - Rocq version: 9.0.0 *)

From mathcomp Require Import interval_inference.
Local Open Scope ring_scope.
Local Open Scope order_scope.
Local Open Scope tensor_scope.

Parameter R : realType.

Parameter f : 'nT[R]_[1] -> 'nT[R]_[1].

Axiom test : forall x, const_t (0 : R) < x ^^ 0 -> const_t (0 : R) < (f x) ^^ 0.
```
Note the missing `From vehicle Require Import utils.` line that is needed

lstrsrmn:
This change fixes the issue I pointed out before:

```diff
diff --git a/vehicle/src/Vehicle/Backend/ITP/Rocq.hs b/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
index 69367d2a..f5e6b11d 100644
--- a/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
+++ b/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
@@ -789,7 +789,7 @@ compileComparison domain op = do
           else ("$0 " <> opDoc <> " $1", dependencies)
   compileNotationAndArgs (dependencies' <> typeDeps) NotAssociative (Just 70) opDoc' Nothing
   where
-    orderDeps = [MathcompImport Boot, Open OrderScope]
+    orderDeps = [VehicleImport VehicleUtils, MathcompImport Boot, Open OrderScope]
     eqDeps = [MathcompImport Boot]
 
 compileStack :: (MonadRocqCompile m) => [Arg DecidabilityBuiltin] -> m Code
```

vehicle/src/Vehicle/Backend/ITP/Rocq.hs
lstrsrmn:
```suggestion
    orderDeps = [VehicleImport VehicleUtils, MathcompImport Boot, Open OrderScope]
```

MatthewDaggitt (approved):

MatthewDaggitt:
Looks like there are minor formatting errors and tests that need to be accepted, but once those are fixed, I'm happy for this to be merged in when you two are!

PR #1131, opened 2026-05-12
feat: add transpose tensor operator

ggustavs:
Adds transpose operator following NumPy semantics, where the dimensions get reversed.


vehicle/src/Vehicle/Compile/Sugar/Resugar/External.hs
```diff
@@ -275,6 +275,7 @@ delabBuiltinFunction fun args = case fun of
   V.StackTensor {} -> rawDelab
   V.ConstTensor -> delabApp (B.Const tokConst) args
   V.Iterate -> rawDelab
+  V.Transpose -> delabTranspose args
```
MatthewDaggitt: This shouldn't need it's own special function. Can't we do this via `delabApp`?

vehicle/src/Vehicle/Data/Builtin/Core.hs
```diff
@@ -124,6 +124,8 @@ data BuiltinFunction
   | -- List operations
     FoldList
   | MapList
+  | -- Tensor shape operations
+    Transpose
```
MatthewDaggitt: Should just go under tensor operations?

vehicle/src/Vehicle/Data/Builtin/Decidability.hs
```diff
@@ -109,6 +109,7 @@ data DecidabilityBuiltin
   | StandardBuiltinFunction BuiltinFunction
   | StandardBuiltinConstructor BuiltinConstructor
   | StandardBuiltinDerivedFunction DerivedFunction
+  | StandardBuiltinCast BuiltinCast
```
MatthewDaggitt: This looks very out of place. Why are we adding casts to the `Decidability` builtin??? They should all have been eliminated by this stage of the compiler.

docs/language/tensors.rst
```diff
@@ -119,6 +119,29 @@ The following operations over tensors are currently supported:
        elements ``A``.
 
 
+Transpose
+---------
+
+``transpose t`` reverses the dimension order of a tensor:
+``Tensor A [d_1, ..., d_n]`` becomes ``Tensor A [d_n, ..., d_1]``.
+For a 2-tensor this is the standard matrix transpose.
```
MatthewDaggitt: Can we just add this to the table above rather than having its own section?

docs/language/tensors.rst
```diff
@@ -119,6 +119,29 @@ The following operations over tensors are currently supported:
...
+(``transpose [[1.0, 2.0], [3.0, 4.0]]``) are folded directly.
+
+Backend support
+~~~~~~~~~~~~~~~
+
+- Verifier (Marabou / VNN-LIB): any rank.
+- Loss: any rank.
+- Agda, Isabelle ITP: any rank.
+- Imandra ITP: any rank, for ``Real``, ``Bool`` and ``Nat`` element types.
+- Rocq ITP: rank 2 only. Targeting Rocq with a higher-rank transpose is an
```
MatthewDaggitt: Would make more sense to add this to the existing limitations section in the ITP documentation (https://vehicle-lang.readthedocs.io/en/stable/exporting.html). Perhaps we need to turn it into a table with features as rows and ITPs as columns with footnotes to explain in detail?

vehicle-rocq/tensor.v
```diff
@@ -759,3 +759,26 @@ by rewrite nstackE ostack_eqE mxE.
...
+   for now the Vehicle compiler emits transpose only at rank 2. *)
+Context {R : Type} {n m : nat}.
+
+Definition transpose_t (t : 'nT[R]_([:: n.+1; m.+1])) : 'nT[R]_([:: m.+1; n.+1]) :=
+  \tensor^^j \tensor^^i t^^i^^j.
+
+(* Correctness: `transpose_t` swaps the two axes — `(transpose_t t)[j][i] = t[i][j]`. *)
+Lemma transpose_tE (t : 'nT[R]_([:: n.+1; m.+1])) (i : 'I_n.+1) (j : 'I_m.+1) :
+  (transpose_t t)^^j^^i = t^^i^^j.
+Proof. by rewrite /transpose_t !nstackE. Qed.
```
MatthewDaggitt: @hoheinzollern is there an existing implementation of transpose in your MathComp PR?

vehicle/src/Vehicle/Backend/ITP/Isabelle.hs
```diff
@@ -55,9 +57,39 @@ data IsabelleOptions = IsabelleOptions
 currentPhase :: Doc ()
 currentPhase = "compilation to Isabelle"
 
+-- | Inline zero-binder non-`@property` `FunctionDecl`s into their use sites
+-- and drop them.
+inlineNonPropertyDecls :: [Decl DecidabilityBuiltin] -> [Decl DecidabilityBuiltin]
+inlineNonPropertyDecls = go Map.empty
```
MatthewDaggitt: Why are we making this change??

vehicle/src/Vehicle/Backend/ITP/Isabelle.hs
```diff
@@ -55,9 +57,39 @@ data IsabelleOptions = IsabelleOptions
...
+      FreeVarUpdate Identity DecidabilityBuiltin
+    replaceFromEnv env recGo p ident args = do
+      args' <- traverse (traverse recGo) args
+      pure $ case Map.lookup ident env of
+        Just body -> normAppList body args'
+        Nothing -> normAppList (FreeVar p ident) args'
+
 compileProgToIsabelle :: (MonadCompile m) => Prog DecidabilityBuiltin -> IsabelleOptions -> m (Doc a)
-compileProgToIsabelle (Main ds) options =
+compileProgToIsabelle (Main ds0) options =
```
MatthewDaggitt: Again, weird change?

vehicle/src/Vehicle/Compile/Unblock.hs
```diff
@@ -212,6 +212,7 @@ unblockRatTensorValue actions@UnblockingActions {..} status expr = do
     VRatStackTensor args -> unblockStackTensor (unblock DifferentDimensions) args
     VRatAt args -> unblockAtTensor (unblock DifferentDimensions) args
     VRatForeach args -> unblockForeachTensor args
+    VRatTensorTranspose {} -> return expr
```
MatthewDaggitt: We need an `unblockTranspose` function here that recursively unblocks it's tensor argument, see e.g. `unblockReduceTensor`. That will remove the need for all the extra code below.

vehicle/src/Vehicle/Compile/Unblock.hs
```diff
@@ -332,7 +333,44 @@ unblockAtTensor unblock (AtTensorArgs tElem d ds xs i) = do
   liftIf xs' $ \xs'' ->
     liftIf i' $ \i'' -> do
       nameCtx <- getNameContext
-      evalAtTensor nameCtx evalApp eval $ AtTensorArgs tElem d ds xs'' i''
+      let unswappedArgs = AtTensorArgs tElem d ds xs'' i''
```
MatthewDaggitt: This function should be unchanged, and every change below it removed.

vehicle/src/Vehicle/Data/Builtin/Standard/Normalise.hs
```diff
@@ -141,6 +144,45 @@ evalVectorToList args@(VectorToListArgs t d xs) =
...
+evalTranspose :: (MonadNormBuiltin m) => EvalSimple TransposeArgs Value Builtin m
+evalTranspose args@(TransposeArgs _ resultDims tensor) =
+  return $
+    fromMaybe (mkExpr accessTranspose args) $
+      -- ConstTensor is uniform: only dims change.
+      goConst <|> goStack2D
+  where
+    goConst :: Maybe (Value Builtin)
+    goConst = case getExpr accessConstTensor tensor of
+      Just (ConstTensorArgs t v _) -> Just $ mkExpr accessConstTensor (ConstTensorArgs t v resultDims)
```
MatthewDaggitt: This implementation looks wrong in several ways... For example, it doesn't take into account tensor literals, and it doesn't correctly reverse the dimensions for the `const` case. I'm not even sure what is going on in the stack case...

I think it will need rewriting from scratch.

vehicle/src/Vehicle/Data/Builtin/Standard/Normalise.hs
```diff
@@ -141,6 +144,45 @@ evalVectorToList args@(VectorToListArgs t d xs) =
...
+    -- 2-D Stack of Stacks: rebuild rows by swapping indices.
+    goStack2D :: Maybe (Value Builtin)
+    goStack2D = case getExpr accessStackTensor tensor of
+      Just (StackTensorArgs t outerDim _ rows) -> do
+        innerStacks <- traverse (getExpr accessStackTensor) rows
+        case innerStacks of
+          [] -> Nothing
+          firstStack@(StackTensorArgs _ innerDim innerRest _) : _ ->
+            -- Only handle 2-D for now: the inner stacks must have empty
+            -- remaining dims.
```
MatthewDaggitt: This is a not a great undocumented limitation. We need to handle arbitrary cases.

ChangeLog.md
```diff
@@ -2,6 +2,16 @@
 
 ## Next release
 
+### Language
+
+* **Feature:** `transpose` operator on tensors. Reverses the
```
MatthewDaggitt: This needs rewriting to be clearer and remove all the extraneous information.

vehicle-imandra/tensor.iml
```diff
@@ -195,6 +195,33 @@ let tensor_from_lookup (ds : int list) (e : int list -> 'a) : 'a tensor =
...
+  | x :: xs' -> list_rev_acc (x :: acc) xs'
+
+let list_rev (xs : 'a list) : 'a list = list_rev_acc [] xs
+
+(* Tensor transpose (real-typed) *)
+
+(* Transpose: reverse dims and lookup with reversed indices.
+  `Option.value` uses a type-appropriate default for the unreachable None. *)
+
+let tensor_transpose_with (default : 'a) (t : 'a tensor) : 'a tensor =
```
MatthewDaggitt: This does not look like the right type signature. Transpose should not need a default value (it should error?). Therefore, there should not need to be three different transpose definitions, and the corresponding complexity in the Imandra backend Haskell code should disappear.

vehicle/src/Vehicle/Backend/ITP/Imandra.hs
```diff
@@ -732,12 +734,48 @@ compileApp :: (MonadImandraCompile m) => Bool -> [ModuleDef] -> Expr Decidabilit
 compileApp _isOutType moduleDefs fun args = do
   let userArgs = NonEmpty.filter (not . wasInsertedByCompiler) args
   case fun of
+    -- Transpose's element type comes in as a compiler-inserted implicit
+    -- arg; we need it to pick the IL monomorphic wrapper, so consult the
+    -- raw args before filtering.
+    Builtin _p b@(StandardBuiltinFunction Transpose) ->
```
MatthewDaggitt: All complexity in this file should disappear once we have the correct implementation in the backend library.

vehicle/tests/golden/features/arithmetic/isabelle/Isabelle.thy.golden
```diff
@@ -24,14 +24,11 @@ begin
...
 
-  definition natLitNatLitDiv  :: "  R FlexTensor "
-    where " natLitNatLitDiv   = ( (pointwise_div (flextensor_from_vec [] [ (((1 :: R) )) ]) (flextensor_from_vec [] [ (((2 :: R) )) ])) ) "
-
   locale  Isabelle  = 
 
     fixes  f  :: " R tensor \<Rightarrow> R FlexTensor "
 
-    assumes  property :  " (geqTensorReduced (flex_subtensor ((f ((combine_subtensors [natLitNatLitDiv])))) (Abs_FlexIndex 0)) (flextensor_from_vec [] [ (((0 :: R) )) ])) "
+    assumes  property :  " (geqTensorReduced (flex_subtensor ((f ((combine_subtensors [(pointwise_div (flextensor_from_vec [] [ (((1 :: R) )) ]) (flextensor_from_vec [] [ (((2 :: R) )) ]))])))) (Abs_FlexIndex 0)) (flextensor_from_vec [] [ (((0 :: R) )) ])) "
```
MatthewDaggitt: Why are we editing so many `Isabelle` tests where the underlying specification hasn't changed?

vehicle/tests/golden/features/transpose/spec2d.vcl
```diff
@@ -0,0 +1,16 @@
+-- Const-free rank-2 transpose: keeps the `transpose` builtin alive into the
+-- ITP backends (the tensor is a network output, so it cannot be folded away)
+-- while avoiding the unrelated `ConstTensor`-ITP bug that `spec.vcl`'s
```
MatthewDaggitt: What is this bug that is spoken of? If there is a bug it should be filed as an `Issue` in the repo.

vehicle/tests/golden/features/transpose/spec3d.vcl
```diff
@@ -0,0 +1,16 @@
+-- Rank-3 transpose through the verifier: exercises the depth-n
+-- index-through-transpose rewrite (regression for the rank>=3 hang).
```
MatthewDaggitt: Again this comment isn't very useful....

vehicle/tests/golden/features/transpose/test.json
```diff
@@ -0,0 +1,95 @@
+[
+  {
+    "name": "TypeCheck",
+    "run": "vehicle typecheck -s spec.vcl",
+    "needs": ["spec.vcl"]
```
MatthewDaggitt: Can you refactor these tests so we have only one spec per directory?

vehicle/tests/golden/specifications/acasXu/isabelle/Isabelle.thy.golden
```diff
@@ -53,21 +50,6 @@ begin
     ultimately show ?thesis by simp
   qed
 
-  definition distanceToIntruder  :: "  FlexIndex "
-    where " distanceToIntruder   = ( (Abs_FlexIndex 0) ) "
```
MatthewDaggitt: Again why have these tests changed so much?

vehicle/tests/golden/features/transpose/IsabelleVerify2d.out.golden
```diff
@@ -0,0 +1 @@
+0
```
MatthewDaggitt, line 1: What is this file?

vehicle/src/Vehicle/Data/Builtin/Decidability/Type.hs
```diff
@@ -201,12 +212,23 @@ convertToDecidabilityFreeVars f p ident args = do
...
   return $ normAppList (FreeVar p ident) finalArgs
   where
+    -- Walk the decl's type. For each leading compiler-inserted implicit Pi,
+    -- prepend a `Hole` arg — UNLESS the call already provides an implicit at
+    -- that position, in which case consume it. This second case matters for
+    -- stdlib decls referenced from stdlib bodies: e.g. `Definitions.append`'s
+    -- type is auto-generalised by the elaborator, marking its `{A}` binder as
+    -- compiler-inserted. Calls in user code don't provide that implicit (so a
+    -- Hole is needed), but calls inside `Definitions.reverse`'s body do (so
+    -- the existing arg fills it).
```
MatthewDaggitt: This description doesn't sound right at all. Whether it was user code or not, after it comes out the type-checker all implicits should be fully inserted. This check isn't about user vs compiler code, but instead seeks to remove implicits that are not needed in the `Decidability` backend.

vehicle/src/Vehicle/Data/Builtin/Interface/Type.hs
```diff
@@ -90,6 +93,22 @@ typeOfBuiltinFunction = \case
   ForeachTensor -> typeOfForeachTensor
   ForeachVector -> typeOfForeachVector
   Iterate -> forAllTypes $ \t -> ((t ~> t) ~> t ~> t) ~> tNat ~> t
+  Transpose -> typeOfTranspose
+
+-- | Stdlib decls that a builtin's type signature refers to. These references
+-- live in compiler code (via `standardLib` in the DSL), so they're invisible
+-- to the AST dependency graph and would otherwise be pruned. Callers that
+-- prune unused decls before secondary-subsystem typecheck need to keep these
+-- as extra roots.
```
MatthewDaggitt: This doesn't seem like the right approach. If the AST pruning code really doesn't take into account these dependencies, the AST pruning algorithm should be updated, rather an adhoc hacks like this...

vehicle/tests/golden/features/transpose/spec.vcl
```diff
@@ -0,0 +1,21 @@
+@network
+f : Tensor Real [2] -> Tensor Real [4, 2]
```
MatthewDaggitt: Could we make this a `[3,2]` tensor instead? Makes it easier to debug...

vehicle/src/Vehicle/Compile/Type/Subsystem.hs
```diff
@@ -73,15 +76,35 @@ decidabilityTypeCheck ::
...
   monoDecProg <- monomorphise decProg isUserCode
   resolveInstanceArgumentsAndCasts monoDecProg
 
+-- | Collect every stdlib decl referenced by some `Builtin`'s type signature
+-- in the prog. These references live in compiler code (see `reverseDims`) so
+-- they don't appear in the AST dependency graph; pass the result as extra
+-- roots when pruning, otherwise the kept stdlib decl is removed and
+-- secondary-subsystem typecheck fails to resolve the FreeVar.
+stdlibTypeLevelDepsOf :: Prog Builtin -> Set Identifier
+stdlibTypeLevelDepsOf prog = execWriter (traverse_ go prog)
```
MatthewDaggitt: Again, if this is a problem, we should be updating the pruning algorithm, not adding hacks like this.

vehicle/src/Vehicle/Compile/Dependency.hs
```diff
@@ -123,7 +124,21 @@ pruneUnusedDeclarations ::
...
+-- `Transpose` builtin's type uses `reverseDims ds` (which expands to a free-var
+-- reference to `Definitions.reverse`). That reference is invisible to
+-- 'createDependencyGraph' (it lives in compiler code, not in any AST node), so
+-- the caller must list such decls explicitly when they need to survive pruning.
+pruneUnusedDeclarationsKeeping ::
+  (MonadCompile m) =>
+  Set Identifier ->
+  Prog builtin ->
+  m (Prog builtin)
+pruneUnusedDeclarationsKeeping extraRoots prog@(Main decls) = do
```
MatthewDaggitt: Again, update the actual pruning algorithm, rather than adding in extraRoots.

vehicle/tests/golden/features/transpose/Marabou.queries/boundedScalar-query1.txt.golden
```diff
@@ -0,0 +1,13 @@
...
+//  - Vehicle version: 0.24.1+dev
+//
+// Assertions
+y0 >= 0.0
+y2 >= 0.0
+y4 >= 0.0
+y6 >= 0.0
+// Input bounds
+x0 = 0.0
+x1 = 0.0
```
MatthewDaggitt: Have you checked that these queries look right?

Also could you generate VNNLIB 2.0 queries for the tests rather than Marabou? We're trying to move away from the Marabou format imminently.

MatthewDaggitt (changes requested):

MatthewDaggitt:
@ggustavs let me know when I should re-review.

vehicle-rocq/tensor.v
hoheinzollern: No, I could add that

hoheinzollern:
@ggustavs @MatthewDaggitt I pushed a rebase to a separate branch adding the Rocq transpose implementation: https://github.com/hoheinzollern/vehicle/commit/9f761c5e43cb8b31997b5db482cffb892b480f32
Since the rebase is substantial I have not force pushed onto here. It needs a thorough review before it gets merged here.

docs/language/tensors.rst
```diff
@@ -229,6 +234,16 @@ The following operations over tensors are currently supported:
...
+   * - Transpose
+     - ``transpose t``
+     - | ``Tensor A ds →``
+       | ``Tensor A (reverse ds)``
+     - | |backendloss_full|
+       | |backendverification_full|
+       | |backendagda_full|
+       | |backendrocq_easy|
+       | |backendimandra_full|
+       | |backendisabelle_full|
```
MatthewDaggitt, line 241: Seems to have been added twice to this file?

docs/exporting.rst
```diff
@@ -73,6 +73,31 @@ support for new ones should be relatively simple, assuming that they have
 the ability to call out to external solvers. Please get in touch if you are
 interested in adding support for a new ITP.
 
+Feature support across backends
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
MatthewDaggitt: Could we rename this heading to `Rocq Limitations` and then just have the short paragraph below. `tensors.rst` should then link to it in the same way that the restrictions on e.g. multiplication link to the verifier page?

vehicle-agda/src/Vehicle/Data/Tensor.agda
```diff
@@ -108,6 +108,17 @@ infix 6 _!_
 _!_ : Tensor A (d ∷ ds) → Fin d → Tensor A ds
 tensor xs ! i = tensor (λ j → xs (combine i j))
 
+-- | Transpose reverses axes: `Tensor A ds -> Tensor A (reverse ds)`.
+-- Implemented by accumulator recursion over dimensions.
+transposeAcc : ∀ {xs ys : Dimensions} {A : Set a}
```
MatthewDaggitt: Can we use `ds` for dimensions rather than `xs` and `ys` which are usually used for the containers themselves in Agda? We don't need the {A : Set a} as A is already declared as a variable.

vehicle-python/src/vehicle_lang/loss/_python/_translation.py
```diff
@@ -342,30 +336,11 @@ def translate_SearchRatTensor(self, expression: vcl.SearchRatTensor) -> py.expr:
...
-                            py_name("len", provenance=vcl.MISSING),
-                            samples_var,
-                            provenance=vcl.MISSING,
-                        ),
-                        provenance=vcl.MISSING,
-                    ),
-                    samples_var,
-                    provenance=vcl.MISSING,
-                ),
-                **asdict(vcl.MISSING),
```
MatthewDaggitt, line 368: I don't understand why a `transpose` PR is altering the implementation of translation of quantifiers??

vehicle-python/src/vehicle_lang/loss/_pytorch/_builtins.py
```diff
@@ -154,6 +154,10 @@ def DenseTensor(
         float_values = [float(val) for val in values]
         return _torch_tensor(data=float_values, dtype=self.dtype_rat).reshape(shape)
 
+    @override
+    def Transpose(self, xs: torch.Tensor) -> torch.Tensor:
+        return xs.permute(*reversed(range(xs.ndim)))
```
MatthewDaggitt, line 160: Can we add a comment here about why were are not use `torch.transpose`? I think it would also be better style to use `torch.permute` as we do in the other operator translations rather than `xs.permute`...

vehicle-syntax/src/Vehicle/Syntax/External.cf
```diff
@@ -249,6 +248,7 @@ If.      Expr4  ::= TokIf Expr5 TokThen Expr5 TokElse Expr5;
 Impl.    Expr5  ::= Expr6 TokImpl Expr5;
 And.     Expr7  ::= Expr8 TokAnd Expr7;
 Or.      Expr7  ::= Expr8 TokOr  Expr7;
+Transpose. Expr14 ::= TokTranspose Expr15;
```
MatthewDaggitt: This seems like a really poor location to put this! Why not with the other tensor operations? e.g. const and stack?

vehicle-syntax/src/Vehicle/Syntax/External.cf
```diff
@@ -358,6 +358,9 @@ DefFunType.   Decl ::= Name TokElemOf Expr;
 DefFunExpr.   Decl ::= Name [NameBinder] "=" Expr;
 DefRecord.    Decl ::= "record" Name [NameBinder] "where" "{" [RecordFieldDef] "}" RecordSupports;
 
+Supports.   RecordSupports ::= "supports" [Name];
+NoSupports. RecordSupports ::= "";
```
MatthewDaggitt, line 366: I'm confused, why is this PR adding this in again? We already have this in `dev`.

MatthewDaggitt (changes requested):
I'm halting the review as it looks like we still haven't got the commit history right on this PR yet?

docs/language/tensors.rst
ggustavs: oops :|

vehicle-python/src/vehicle_lang/loss/_python/_translation.py
ggustavs: I can separate it out to another branch, but I was running into cases where lambdas were reaching tensorflow/ pytorch tensor operations.

vehicle-python/src/vehicle_lang/loss/_pytorch/_builtins.py
ggustavs: It is because torch.transpose swaps exactly two dimensions, while torch.permute can be used to reorder all dimensions. If you think that should be added as a comment, I can add one that points to torch docs :)

vehicle-syntax/src/Vehicle/Syntax/External.cf
ggustavs: It is moving them to where they are used, since the dev version defines them ~150 lines above where it is used. If you don't think that's a change that should be made here, I can revert it :)

MatthewDaggitt: Ah okay, apologies I missed that!

ChangeLog.md
MatthewDaggitt, line 3: I'm sorry I forgot to update the changelog with the v0.26.0 release. The existing contents needs to be put under `## Version 2.0` and a new `Next release` section needs to be created with your new changes listed there.

vehicle/src/Vehicle/Data/Tensor.hs
```diff
@@ -205,6 +206,19 @@ extendTensor dim = \case
   ConstantTensor shape value -> ConstantTensor (dim : shape) value
   DenseTensor shape values -> DenseTensor (dim : shape) (Vector.concat (replicate dim values))
 
+transposeTensor :: (Eq a) => Tensor a -> Tensor a
+transposeTensor = \case
+  ConstantTensor shape value -> ConstantTensor (reverse shape) value
+  DenseTensor shape values ->
+    let revShape = reverse shape
+        pickAt revIs = values Vector.! flattenIndices shape (reverse revIs)
```
MatthewDaggitt: Can we use the standard `do` notation here with `let`s on each line and no `in`?

vehicle/src/Vehicle/Data/Code/Interface/Args.hs
```diff
@@ -327,6 +327,25 @@ instance IsArgs ForeachTensorArgs where
...
+    transposeTensor :: expr
+  }
+
+instance IsArgs TransposeArgs where
+  accessSpine =
+    Access
+      { getExpr = \case
+          (fmap argExpr -> [t, ds, xs]) -> Just $ TransposeArgs t ds xs
+          _ -> Nothing,
+        mkExpr = \(TransposeArgs t ds xs) -> [implicit t, implicitIrrelevant ds, explicit xs]
```
MatthewDaggitt: I don't think the dimensions should be irrelevant here and the normalisation procedure relies on them? (if you don't know the dimensions, you can't transpose the tensor).

vehicle/src/Vehicle/Backend/ITP/Isabelle.hs
```diff
@@ -56,25 +56,63 @@ data IsabelleOptions = IsabelleOptions
 currentPhase :: Doc ()
 currentPhase = "compilation to Isabelle"
 
+freeVarRefs :: Expr DecidabilityBuiltin -> Set Identifier
+freeVarRefs = execWriter . go
```
MatthewDaggitt: Why do we have to make such large scale changes to the Isabelle compilation file to add another tensor operation?

vehicle/src/Vehicle/Data/Builtin/Interface/Blocked.hs
```diff
@@ -86,6 +86,7 @@ functionBlockingStatus b spine = case b of
   ForeachTensor -> fixedStatus [1] spine
   ForeachVector -> fixedStatus [1] spine
   Iterate -> fixedStatus [2] spine
+  Transpose -> DoesNotReduce
```
MatthewDaggitt: This seems wrong? We've defined a normalisation procedure so it clearly should reduce?

vehicle/src/Vehicle/Data/Builtin/Interface/Normalise.hs
```diff
@@ -571,6 +572,36 @@ evalAtTensor ctx evalApp eval args@(AtTensorArgs t d ds tensor index) =
         evalApp ctx fn [explicit index]
       _ -> Nothing
 
+    -- `(transpose t) ! i1 ! ... ! in` (scalar result) = `t ! in ! ... ! i1`.
```
MatthewDaggitt: This equality makes no sense to me. At this point in the code we know we are matching on a term of the form `transpose xs ! i`, but we certainly don't have the full chain of indices as suggested by that equation, so I really don't think there is any simplification that we can do here. What is the code below trying to achieve?

vehicle/src/Vehicle/Data/Builtin/Standard/Normalise.hs
```diff
@@ -144,6 +150,84 @@ evalVectorToList args@(VectorToListArgs t d xs) =
     INatLiteral n | n == length xs -> mkListExpr (argExpr t) xs
     _ -> mkExpr accessFromVectorToList args
 
+evalTranspose :: (MonadNormBuiltin m) => EvalSimple TransposeArgs Value Builtin m
```
MatthewDaggitt: This needs to go in `Builtin.Interface.Normalise` and be generalised so that it works with any `builtin` type (e.g. see `evalAt` for an example). We should then be able to use it with `LossBuiltin` see comment.

vehicle/src/Vehicle/Data/Builtin/Standard/Normalise.hs
```diff
@@ -144,6 +150,84 @@ evalVectorToList args@(VectorToListArgs t d xs) =
...
+       in mkExpr
+            accessStackTensor
+            ( StackTensorArgs
+                tElem
+                (INatLiteral d)
+                (foldr (IDimCons . INatLiteral) IDimNil ds)
+                subStacks
+            )
+
+    chunksOf :: Int -> [a] -> [[a]]
```
MatthewDaggitt: We already have a utility function for this available in `Data.List.Split`

vehicle/src/Vehicle/Data/Builtin/Loss.hs
```diff
@@ -353,6 +356,7 @@ instance NormalisableBuiltin LossBuiltin where
       ConstTensor -> Simple evalConstTensor
       FoldList -> NonSimple evalFoldList
       MapList -> NonSimple evalMapList
+      Transpose -> None
```
MatthewDaggitt: Once generalised, we should be able to use `evalTranspose` here (see comment above).

vehicle/src/Vehicle/Compile/Type/Subsystem.hs
```diff
@@ -73,8 +74,9 @@ decidabilityTypeCheck ::
   Prog Builtin ->
   m (Prog DecidabilityBuiltin)
 decidabilityTypeCheck prog = do
-  prunedProg <- pruneUnusedDeclarations prog
-  errorOrDecProg <- typeCheckWithSubsystem DecidabilityTypes decidabilityBuiltinInstances prunedProg
+  prunedProg <- pruneUnusedDeclarations standardBuiltinTypeDeps prog
+  castFreeProg <- resolveInstanceArgumentsAndCasts prunedProg
```
MatthewDaggitt: Why does adding `transpose` require us to do an extra round of `resolveInstanceArgumentsAndCasts`? This should already have been done after the spec was type-checked with the main type system?

vehicle/src/Vehicle/Data/Builtin/Decidability/Type.hs
```diff
@@ -200,12 +200,18 @@ convertToDecidabilityFreeVars f p ident args = do
...
     insertNewArgs :: [Arg DecidabilityBuiltin] -> Type DecidabilityBuiltin -> m [Arg DecidabilityBuiltin]
     insertNewArgs as = \case
-      Pi _ binder result -> do
-        if wasInsertedByCompiler binder && isImplicit binder
-          then (argFromBinder binder (Hole p "_") :) <$> insertNewArgs as result
-          else return as
+      Pi _ binder result | wasInsertedByCompiler binder && isImplicit binder ->
+        case as of
+          a : rest
+            | isImplicit a ->
```
MatthewDaggitt: This doesn't look right to me... Whether or not the arg iteself is present and implicit, is not relevent to whether it needs to be replaced in the `Decidability` type system. Indeed as the specification is well-typed at this point, this branch should always evaluate to true so what's the point in including it?

vehicle/src/Vehicle/Compile/Dependency.hs
```diff
@@ -83,8 +83,12 @@ fromEdges (AdjacencyGraph outEdges) = do
 --------------------------------------------------------------------------------
 -- Constructing the dependency graph
 
-createDependencyGraph :: [Decl builtin] -> DependencyGraph Identifier
-createDependencyGraph ds = fromEdges $ AdjacencyGraph $ Map.fromList $ fmap goDecl ds
+createDependencyGraph ::
+  forall builtin.
+  (builtin -> Set Identifier) ->
```
MatthewDaggitt: Instead of passing this through as a parameter everywhere, could we add it as a method on the type class to the `TypableBuiltin` and then add that as a constraint here? I think it localises it a little bit better as well as ensuring that it works for other builtins as well.

MatthewDaggitt (changes requested):
Thanks, have left the reminder of the review!

MatthewDaggitt:
I will fix the merge conflicts for this over the next few days.

MatthewDaggitt (approved):

MatthewDaggitt:
Okay, I think I'm happy with this now so merging in.

If we need to fix the search translation, then we should do it as separate PR :+1:

PR #1132, opened 2026-05-13
feat: compile non-property decls to loss

ggustavs:
Add ability to compile non-property declarations to callable functions translated with the chosen DL. This is done by passing the name of the declaration via the `--declaration` flag, same as for properties.


ChangeLog.md
```diff
@@ -4,6 +4,8 @@
 
 ### Loss backend
 
+* Declaration flag can now be used to specify non-property declarations that will also be compiled with the chosed DL to a callable function.
```
MatthewDaggitt: - The `--declaration` flag. 
- "chosed" -> "chosen"
- "to a callable function"? Is this true if the declaration is `x : Int; x = 2`? That shouldn't be compiled to a callable right?

vehicle/src/Vehicle/Backend/Loss.hs
```diff
@@ -42,31 +47,34 @@ convertDecls ::
...
+convertDecl logicID logic requestedDecls decl = do
   logCompilerSection2 MinDetail ("declaration" <+> quotePretty (identifierOf decl)) $ do
     runMonadLogicT logicID logic decl $ do
       case decl of
         DefAbstract p ident sort typ
           | isExternalResourceDecl decl -> Just <$> convertResourceDecl p ident sort typ
           | otherwise -> return Nothing
         DefFunction p ident ann typ expr
-          | isPropertyDecl decl -> Just <$> convertPropertyDecl p ident ann typ expr
+          | isPropertyDecl decl || Set.member (nameOf ident) requestedDecls ->
```
MatthewDaggitt: Hmm should we be changing the semantics so that it only compiles the requested declarations? This seems a little odd to compile both, even if the user has only requested specific, non-property declarations?

vehicle/src/Vehicle/Compile.hs
```diff
@@ -154,7 +155,7 @@ compileToLossFunction ::
   m ()
 compileToLossFunction LossOptions {..} typedProg outputAsJSON =
   logCompilerPass Loss $ do
-    lossTensorProg <- convertToLossTensors differentiableLogicID typedProg
+    lossTensorProg <- convertToLossTensors differentiableLogicID (Set.fromList declarationsToCompile) typedProg
```
MatthewDaggitt: Can we put the declaration set calculation on a separate line?

vehicle/tests/golden/errors/lossOutputs/unsupportedType/Loss.err.golden
```diff
@@ -0,0 +1,2 @@
+Error in file 'spec.vcl' at Line 4, Columns 1-9: compiling declaration 'someList' as a loss output (supported leaf types: `Bool`, `Real`, `Tensor Bool _`, `Tensor Real _`, `Vector _ _`) is not currently implemented
```
MatthewDaggitt: This limitation feels a bit artificial. It should be possible to get it working for any declaration...

vehicle/src/Vehicle/Backend/Loss.hs
```diff
@@ -83,58 +91,82 @@ convertResourceDecl p ident sort typ = do
...
-convertPropertyDecl p ident ann typ value = do
-  lossType <- convertDeclType typ
-  lossValue <- convertMultiProperty typ value
-  let lossExpr = unnormalise 0 lossValue
-  let lossTensorDecl = DefFunction p ident ann lossType lossExpr
-  return lossTensorDecl
+convertOutputDecl p ident ann typ value = do
+  leafType <- stripPiTypes typ
+  case toTypeValue leafType of
+    VBoolTensorType {} -> emit leafType
```
MatthewDaggitt: This code is completely baffling. Why are we doing the same thing for every type value? We should be doing something different for each one...

vehicle/src/Vehicle/Backend/Loss.hs
```diff
@@ -83,58 +91,82 @@ convertResourceDecl p ident sort typ = do
...
+            <+> "as a loss output (supported leaf types: `Bool`, `Real`, `Tensor Bool _`, `Tensor Real _`, `Vector _ _`)"
+  where
+    emit leafType = do
+      lossType <- convertDeclType typ
+      lossValue <- convertFunction (convertMultiOutput leafType) value
+      let lossExpr = unnormalise 0 lossValue
+      return $ DefFunction p ident ann lossType lossExpr
+
+stripPiTypes :: (MonadLogic m) => VType Builtin -> m (VType Builtin)
+stripPiTypes typ = case toTypeValue typ of
```
MatthewDaggitt: This doesn't look right either. Stripping all Pi types off means we can't convert functions... Is this trying to get rid of the implicit functions? If so I don't think they should exist at this point because the whole spec has been monomorphised?

Is there a concrete test case that we need to do this for?

vehicle/src/Vehicle/Backend/Loss.hs
```diff
@@ -83,58 +91,82 @@ convertResourceDecl p ident sort typ = do
...
 convertDeclType :: (MonadLogic m) => VType Builtin -> m (Type LossBuiltin)
 convertDeclType typ = unnormalise 0 <$> convertType typ
 
-convertMultiProperty :: (MonadLogic m) => VType Builtin -> Value Builtin -> m (Value LossBuiltin)
-convertMultiProperty typ = case toTypeValue typ of
-  VBoolTensorType _ds -> convertTensorProperty
-  VVectorType tElem _d -> convertVectorProperty tElem
-  _ -> unexpectedExprError currentPass "Impossible property type"
+convertMultiOutput :: (MonadLogic m) => VType Builtin -> Value Builtin -> m (Value LossBuiltin)
+convertMultiOutput typ = case toTypeValue typ of
```
MatthewDaggitt: Can we rename this simply `convertTypedValue`?

vehicle/src/Vehicle/Backend/Loss.hs
```diff
@@ -83,58 +91,82 @@ convertResourceDecl p ident sort typ = do
...
+  VBoolTensorType _ds -> convertBoolTensorOutput
+  VRatTensorType _ds -> convertRatTensor
+  VVectorType tElem _d -> convertVectorOutput tElem
+  VBoolType -> convertBoolTensorOutput
+  VRatType -> convertRatTensor
+  _ -> unexpectedExprError currentPass "unsupported leaf type"
 
-convertVectorProperty :: (MonadLogic m) => VType Builtin -> Value Builtin -> m (Value LossBuiltin)
-convertVectorProperty typ value = do
+convertVectorOutput :: (MonadLogic m) => VType Builtin -> Value Builtin -> m (Value LossBuiltin)
```
MatthewDaggitt: Can we rename this simply `convertVectorValue`? (and similar renamings for the other functions).

MatthewDaggitt (changes requested):
Looks like the right idea, thank you! Got a few minor suggestions for changes.

vehicle/src/Vehicle/Backend/Loss.hs
ggustavs: I was thinking more about the context of needing more from a spec, not just the properties, but your approach makes more sense in general

MatthewDaggitt:
Okay, I have brought this up to date and refactored it so that this is a much more minimal change. @ggustavs does it look good to you?

MatthewDaggitt (approved):

PR #1133, opened 2026-05-15
Allow record and record accesses in query compilation

laurenpudz:
Adds support for records and record accesses in query compilation. 
- Adds function to wrap record quantifiers in rat tensor quantifier
- Adds types to typedView for record quantifiers and accesses
- Adds eval and unblocking for record accesses (and unblocking for records)
- Changes network input/output types to a networkIOtype, which is either a tensor or a record
- refactors some existing tensorisable code 
- *still a WIP* - there are some undefineds and places marked `TODO: RECORD SUPPORT` that need to be looked at
- Tests are failing because plan files are now slighly different sizes due to different naming of network types. Happy to change these once I've gotten confirmation that changing them is actually necessary
- Adds query compilation step to record quantifier test


vehicle/src/Vehicle/Data/AST/Provenance.hs
```diff
@@ -116,6 +116,18 @@ instance ToJSON Provenance where
         "contents" .= toJSON @[Int] [posLine start, posColumn start, posLine end, posColumn end]
       ]
 
+instance FromJSON Provenance where
```
MatthewDaggitt: Oh we shouldn't need this instance 😅 If we need it because we are serialising `RecordField` then we should probably just be serialising `FieldName` instead without the provenance information?

vehicle/src/Vehicle/Compile/Normalise/NBE.hs
```diff
@@ -162,6 +164,18 @@ evalRecordDef = go mempty emptyBoundEnv
         fields' <- traverseRecordFields (eval ctx boundEnv) fields
         return ([], fields')
 
+evalRecordAcc ::
+  (MonadNorm builtin m, MonadFreeContext builtin m) =>
+  Value builtin ->
+  FieldName ->
+  m (Value builtin)
+evalRecordAcc value fieldName = do
```
MatthewDaggitt: Can we reuse this on Line 201+?

vehicle/src/Vehicle/Compile/Normalise/NBE.hs
```diff
@@ -162,6 +164,18 @@ evalRecordDef = go mempty emptyBoundEnv
...
+  fields <- case value of
+    VRecord _typ fields -> return fields
+    _ -> developerError "Record declaration is not of expected format."
```
MatthewDaggitt: We should rebuild the record in this case like we do at Line 201+

vehicle/src/Vehicle/Compile/Unblock.hs
```diff
@@ -181,6 +184,20 @@ unblockBoolMultiDimTensorValue actions expr = do
...
 
+unblockRecordValue :: UnblockingActions m -> DimensionsStatus -> UnblockingFunction m
+unblockRecordValue actions@UnblockingActions{..} status expr = do
+  showEntry expr
+  showExit =<< case toRecordValue expr of
+    VRecordFreeVar n spine -> case getExpr accessSpine spine of
+      Just args -> do
+        unblockNetworkApp (unblockTensor status) (unblockRecord status) n args
+      _ -> return expr
+    VRecordBoundVar {} -> return expr
```
MatthewDaggitt: Should probably be using `unblockRecordBoundVar` here?

vehicle/src/Vehicle/Data/Code/TypedView.hs
```diff
@@ -60,19 +64,24 @@ data TypeValue
   | VRatType
   | VBoolTensorType (VDims Builtin)
   | VNatTensorType (VDims Builtin)
-  | VRatTensorType (VDims Builtin)
+  | VTensorLike TensorLikeValue
```
MatthewDaggitt: Please can we just expand this here rather than having it's own data-type?

vehicle/src/Vehicle/Data/Code/TypedView.hs
```diff
@@ -47,11 +50,12 @@ import Vehicle.Data.Variable.Bound.Level
 import Vehicle.Data.Variable.Free.Context (MonadFreeContext)
 import Vehicle.Prelude
 import Vehicle.Prelude.Logging
+import Vehicle.Data.Builtin.Interface.Print
 
 -------------------------------------------------------------------------------
 -- Types
 
--- | A view on all possible expressions that can have type `List Int`.
+-- | A view on all possible expressions that can have type `Type`.
```
MatthewDaggitt, line 57: Nice spot!

vehicle/src/Vehicle/Backend/Solver/UserVariableElimination.hs
```diff
@@ -245,7 +276,8 @@ eliminateTensorAssertion ::
   ComparisonOp ->
   TensorOp2Args (Value Builtin) ->
   m (Value Builtin)
-eliminateTensorAssertion op (TensorOp2Args dims xs ys) =
+eliminateTensorAssertion op (TensorOp2Args dims xs ys) = do
+  _ <- logDebug MidDetail $ "dims are" <+> pretty (show dims) <+> "xs are" <+> pretty (show xs) <+> "ys are" <+> pretty (show ys)
```
MatthewDaggitt: Should remove this.

vehicle/src/Vehicle/Compile/Resource.hs
```diff
@@ -32,26 +33,46 @@ instance Pretty NetworkType where
...
 
-instance ToJSON NetworkTensorType
 
-instance FromJSON NetworkTensorType
+type GenericRecordFieldNames = [FieldName]
+
+tensorSize :: NetworkIOType -> Int
+tensorSize typ = case typ of
+  NetworkTensorType _ dims -> product dims
+  NetworkRecordType _ _ dims _ -> product dims
```
MatthewDaggitt: I think this `tensorSize` function isn't used, so we can just delete?

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -304,3 +304,53 @@ createComparisonField ::
 createComparisonField recordType toTensor fieldIdent = do
   fromDSL mempty $ explLam "r1" recordType $ \r1 ->
     explLam "r2" recordType $ \r2 -> fieldIdent @@ [toTensor @@ [r1], toTensor @@ [r2]]
+
+-- -----------------------------------------------------------------------------------------------
+-- Record/Tensorisable util functions
+-- Not sure if these should go here or if they are at the right level of abstraction
```
MatthewDaggitt, line 309: I would put these 3 functions in the same file where the `MonadFreeContext` file lives? And make them just run in `MonadFreeContext` and take an ident rather than explicitly take a `DeclContextEntry`?

vehicle/src/Vehicle/Data/Builtin/Standard/Scoping.hs
```diff
@@ -304,3 +304,53 @@ createComparisonField ::
...
+  m GenericRecordFieldNames
+getRecordFieldNames r = case r of
+  DefRecord _p _ident _sort _telescope fields -> return $ map fst fields
+  _ -> compilerDeveloperError "Record declaration is not of expected format."
+
+constructFromTensorFreeVar ::
+  Identifier ->
+  Provenance ->
+  Expr Builtin
+constructFromTensorFreeVar ident p =
```
MatthewDaggitt: Use `do` notation and get rid of the `in` in these two functions.

vehicle/src/Vehicle/Compile/Resource.hs
```diff
@@ -32,26 +33,43 @@ instance Pretty NetworkType where
...
-instance Pretty NetworkTensorType where
-  pretty tensor =
-    "Tensor"
-      <+> pretty (baseType tensor)
-      <+> pretty (dimensions tensor)
+instance Pretty NetworkIOType where
+  pretty = \case
+    NetworkTensorType t dims -> "Tensor" <+> pretty t <+> pretty dims
+    NetworkRecordType t ident dims fields ->
+      "Record" <+> pretty ident <+> ":" <+> encloseSep ("{" <> space) (space <> "}") ("," <> space) (map prettyField fields)
```
MatthewDaggitt: Should be able to use `prettyMapEntries` here from `Vehicle.Prelude.PrettyPrinter`. I know it prints an extra `:` but I don't think it's a problem.

vehicle/src/Vehicle/Compile/Resource.hs
```diff
@@ -14,8 +15,8 @@ import Vehicle.Prelude
 type NetworkName = Name
 
 data NetworkType = NetworkType
-  { inputTensor :: NetworkTensorType,
-    outputTensor :: NetworkTensorType
+  { inputTensor :: NetworkIOType,
```
MatthewDaggitt: Can we rename this fields `networkInputType` and `networkOutputType`?

vehicle/src/Vehicle/Compile/Resource.hs
```diff
@@ -32,26 +33,43 @@ instance Pretty NetworkType where
...
-instance NFData NetworkTensorType
+instance NFData NetworkIOType
+
+instance ToJSON NetworkIOType
 
-instance ToJSON NetworkTensorType
+instance FromJSON NetworkIOType
 
-instance FromJSON NetworkTensorType
+type GenericRecordFieldNames = [FieldName]
```
MatthewDaggitt: I don't think this type synonym is very useful, let's just go with `[FieldName]`.

vehicle/src/Vehicle/Compile/Resource.hs
```diff
@@ -32,26 +33,43 @@ instance Pretty NetworkType where
...
 networkSize network = tensorSize (inputTensor network) + tensorSize (outputTensor network)
 
-data NetworkTensorType = NetworkTensorType
-  { baseType :: NetworkBaseType,
-    dimensions :: TensorShape
-  }
+data NetworkIOType
+  = NetworkTensorType
+      { baseType :: NetworkBaseType,
+        dimensions :: TensorShape
```
MatthewDaggitt: So unfortunately defining fields like this for multi-constructor datatypes is heavily discouraged as the projection functions only work if the datatype has the right constructor, e.g. calling the function `baseType : NetworkIOType -> NetworkBaseType` will error if the `NetworkIOType` is a `NetworkRecordType`. 

Therefore I think the best course might be to define this as something like the following?
```
data NetworkTensorType = NetworkTensorType
  { baseType :: NetworkBaseType,
    dimensions :: TensorShape
  }

data NetworkRecordType = NetworkRecordType
  { recordTypeIdent :: Identifier,
    recordFields :: [FieldName],
    fieldTensorType :: NetworkTensorType
  }

data NetworkIOType 
  = TensorIOType NetworkTensorType
  | RecordIOType NetworkRecordType
```

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
```diff
@@ -103,8 +103,14 @@ networkTensor :: Name -> NetworkType -> Int -> InputOrOutput -> (Name, TensorSha
...
-    Input -> (name, dimensions $ inputTensor networkType)
-    Output -> (name, dimensions $ outputTensor networkType)
+    Input -> (name, getNetworkIODimensions $ inputTensor networkType)
+    Output -> (name, getNetworkIODimensions $ outputTensor networkType)
+
+getNetworkIODimensions :: NetworkIOType -> TensorShape
+getNetworkIODimensions ioTensor =
+  case ioTensor of
+    NetworkTensorType _ dims -> dims
+    NetworkRecordType _ _ dims _ -> dims
```
MatthewDaggitt: Should this be `length fields : dims`?

vehicle/src/Vehicle/Backend/Solver.hs
```diff
@@ -245,6 +253,67 @@ compileQuantifiedQuerySet isPropertyNegated args =
...
+  recordTypeDecl <- getDeclEntry (Proxy @Builtin) recordTypeIdent
+  shape <- getRecordDimsExpr recordTypeDecl
+  let dims = mkDims shape
+
+  -- Build tensor binder with appropriate dims and type for record
+  let Closure boundEnv _body = quantifyRecordBody
+  tensorType <- eval namedCtx boundEnv $ fromDSL mempty $ tTensor tRat (toDSL dims)
+  normalisedDims <- eval namedCtx boundEnv dims
+
+  let tensorBinder =
```
MatthewDaggitt: You could use `mkExplicitBinder` here.

vehicle/src/Vehicle/Backend/Solver.hs
```diff
@@ -245,6 +253,67 @@ compileQuantifiedQuerySet isPropertyNegated args =
...
+        Binder
+          { binderDisplayForm = BinderDisplayForm (NameAndType (getFreshTensorBinderName namedCtx) mempty) True,
+            binderVisibility = Explicit,
+            binderRelevance = Relevant,
+            binderValue = tensorType
+          }
+
+  let tensorBoundVar = Arg Explicit Relevant (BoundVar mempty 0)
```
MatthewDaggitt: Use `explicit` function from `Data.AST.Arg`?

vehicle/src/Vehicle/Backend/Solver.hs
```diff
@@ -245,6 +253,67 @@ compileQuantifiedQuerySet isPropertyNegated args =
...
+wrapQuantifyRecord ::
+  ( MonadPropertyStructure m,
+    MonadSupply QueryID m,
+    MonadStdIO m,
+    MonadFreeContext Builtin m
+  ) =>
+  QuantifyRecordArgs (Value Builtin) (Closure Builtin) ->
+  m (QuantifyRatTensorArgs (Value Builtin) (Closure Builtin))
+wrapQuantifyRecord QuantifyRecordArgs {..} = do
+  -- Takes a record quantifier and wraps the binder & body in a tensor quantifier
```
MatthewDaggitt: Can we add this as a doc string to the function. i.e. put it above `wrapQuantifyRecord` and start it with `-- |` instead of `--` on the first line.

vehicle/src/Vehicle/Backend/Solver.hs
```diff
@@ -245,6 +253,67 @@ compileQuantifiedQuerySet isPropertyNegated args =
     (maybePartitions, globalCtx) <- runStateT (eliminateExists args) emptyGlobalCtx
     compileQuerySetPartitions globalCtx isPropertyNegated maybePartitions
 
+-- Generates an unused binder name of the form '_tN', where N is an integer
+getFreshTensorBinderName ::
+  NamedBoundCtx ->
+  Text.Text
+getFreshTensorBinderName ctx = checkExistsInCtx 0
```
MatthewDaggitt: Can we put this in one in the files in the `MonadNameContext` subfolder?

MatthewDaggitt (changes requested):
Looks really good! I have a list of hopefully very minor tweaks and tidying up, but nothing major.

vehicle/src/Vehicle/Verify/QueryFormat/VNNLib.hs
laurenpudz: This should be the dims of the tensor equivalent of the record. The `length fields : dims` gets handled by `getRecordDims` when we construct the `NetworkRecordType`. It is unclear though, so I'll add a comment, and am more than happy to change it over if you think it would make more sense the other way.

vehicle/src/Vehicle/Backend/Solver.hs
laurenpudz: I can't find that subfolder, do you think `Compile\Type\Monad.hs` would be ok?

MatthewDaggitt: Sorry, I was being unclear. `Vehicle.Data.Variable.Bound.Context.Name.Core`!

MatthewDaggitt:
You can just accept the tests that change the size of the golden files. That is to be expected if we are changing the network type.

MatthewDaggitt (approved):

PR #1135, opened 2026-05-18
Prevent Vehicle tests from running unnecessarily

MatthewDaggitt:
Trying to prevent the full test-suite from running when we make minor changes elsewhere (e.g. update the documentation).

@wenkokke, @ggustavs does this look okay?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,35 @@ defaults:
     shell: sh
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      build-vehicle: ${{ steps.filter.outputs.build-vehicle }}
+      build-vehicle-python: ${{ steps.filter.outputs.build-vehicle-python }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            build-vehicle:
+              - 'vehicle/**'
+              - 'vehicle-syntax/**'
+            build-vehicle-python:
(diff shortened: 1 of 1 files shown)


wenkokke:
This will cause an issue with the merge conditions, which require that certain jobs ran and completed successfully.

wenkokke:
If you're gonna just make runs conditional on certain files changing, there's no need to involve a whole other action, since GitHub's `on` conditions can ensure workflows only run if certain files are changed.

wenkokke:
However, you have to make sure that every required job runs, service means not skipping them, but rather having them succeed immediately. That'll get a bit messy.

wenkokke:
I'd suggest the following for Vehicle:

- Run `cabal sdist` and compute the sha512 hash of the sdist.
- Compare the hash with hash of sdist in target branch.
- If they're the same, don't run the vehicle tests.

Similar for the Python tests, expect obviously with a different command to build the sdist.

In both cases, you also need to compare the hashes of the actions and workflows.

You can even use the GitHub cache action to store a map from commits to hashes, e.g., as a little JSON file.

That should be way less messy than manually maintaining a list of files. Moreover, you can have it run in each individual job, so that the job simply succeeds without running the tests if they're not needed, so the merge conditions aren't messed up.

All of this logic can probably be a short shell script that uses git to read the current commit hash, cabal/uv to create the sdists, sha512sum for hashing the sdists, and jq for editing the cached commit to sdist hash map.

wenkokke (changes requested):
See comments.

MatthewDaggitt:
Thanks for the feedback @wenkokke! I had forgotten about the inter-job dependencies. More complicated then I realised....

PR #1141, opened 2026-05-24
Fixed Verifier tests and made it so `TestVerifier` tests always enabled

MatthewDaggitt:
Currently we were only running `TestVerifier` tests when it was explicitly passed on the allowlist, but it makes sense to run them all the time so we don't get regressions. Have achieved this by adding it as a default option in the Vehicle test suite declaration.


MatthewDaggitt:
Ah okay, yes, the test verifier doesn't work on Windows so disabling again. It appears we already run the `TestVerifier` tests on Ubuntu CI so we're probably good.

PR #1143, opened 2026-05-24
Use record information in compiler reporting

laurenpudz:
Provides support for record information in counterexample reporting and bounds errors
*note: this is still a WIP*
Changes:
- Adds a new type of compilation step tracking quantified record name, fields, and name of binder it gets wrapped in
- Generates a step of this type with along with the wrapped quantifier args and prepends it on to existing steps
- Reconstructs records for counterexample reporting during user variable reconstruction
- ^ code for this is quite nasty and involves skipping over the relevant quantifier steps during most of reconstruction, so I'm not sure if we should do this a different way
- Adds alternative error message for bounds errors over quantified record var
- ^ having trouble accessing the name of the quantified variable so it is currently just hardcoded as `x`, but this will need to change
- Added tests, for some reason the progress bar for the counterexample test will not match, whatever I do :/


vehicle/tests/golden/errors/tensorLike/tensorLikeCounterexample/Marabou.out.golden
```diff
@@ -0,0 +1,4 @@
+Verifying properties:
+  p [==============================================================] 1/1 queries 0/1 queries
+    result: [91m✗[m - TestVerifier found a counterexample
+      x: { a = 0.2 b = 0.2 }
```
MatthewDaggitt: Can we print this using `prettyMapLike` or one of its friends so that it appears with each field laid out on different lines?

vehicle/tests/golden/errors/tensorLike/tensorLikeMissingBounds/Marabou.err.golden
```diff
@@ -0,0 +1,4 @@
+Error in file 'spec.vcl' at Line 14, Columns 1-2: The property 'User.p' cannot be compiled as cannot deduce lower and upper bounds for the input of 
+  f x
+In particular, missing upper bounds for fields x.a x.b for 'x'
```
MatthewDaggitt: Should be a comma or an and between `x.a` and `x.b`? Also could we get the name of the record type printed here, e.g. "for Pair 'x'"/"for tensor 'x'"

vehicle/src/Vehicle/Verify/Core.hs
```diff
@@ -176,13 +176,28 @@ flipQueryRel = \case
...
-    vsep (fmap pretty assignment)
+data UserVariableAssignmentType 
+  = TensorAssignment (Name, RatTensor)
+  -- Attempting to cheat recordfields here
+  | RecordAssignment (Name, [(Name, RatTensor)])
+  deriving (Show, Generic, Eq)
+
+instance ToJSON UserVariableAssignmentType
+
+instance FromJSON UserVariableAssignmentType
```
MatthewDaggitt: Add empty line. Looks like you haven't got the pre-commit hooks installed? See CONTRIBUTING.md for details.

vehicle/src/Vehicle/Verify/Core.hs
```diff
@@ -176,13 +176,28 @@ flipQueryRel = \case
 
 -- | A (satisfying) assignment to a set of user-level variables.
 newtype UserVariableAssignment
-  = UserVariableAssignment [(Name, RatTensor)]
+  = UserVariableAssignment [UserVariableAssignmentType]
```
MatthewDaggitt: Would make more sense to extract the `Name` field, i.e. `[(Name, UserVariableAssignment)]`

vehicle/src/Vehicle/Verify/Core.hs
```diff
@@ -176,13 +176,28 @@ flipQueryRel = \case
...
+    vsep (fmap pretty assignments)
+
 instance ToJSON UserVariableAssignment
 
 instance FromJSON UserVariableAssignment
 
-instance Pretty UserVariableAssignment where
-  pretty (UserVariableAssignment assignment) =
-    vsep (fmap pretty assignment)
+data UserVariableAssignmentType
```
MatthewDaggitt: Rename to `UserVariableValue`? and constructors to `TensorValue`/`RecordValue`?

vehicle/src/Vehicle/Verify/Core.hs
```diff
@@ -176,13 +176,28 @@ flipQueryRel = \case
...
+  deriving (Show, Generic, Eq)
+
+instance ToJSON UserVariableAssignmentType
+
+instance FromJSON UserVariableAssignmentType
+instance Pretty UserVariableAssignmentType where
+  pretty t = case t of
+    TensorAssignment tens-> pretty tens
+    -- LAUREN TODO: this is probably pretty dodgy
+    RecordAssignment (_varName, fields) -> vsep (fmap (\(name, tens) -> pretty name <+> pretty tens) fields)
```
MatthewDaggitt: Use `prettyMapLike` or something similar.

vehicle/src/Vehicle/Verify/Specification/Execute/Reporting.hs
```diff
@@ -262,8 +261,12 @@ statusSymbol verified = do
 
 prettyUserVariableAssignment :: UserVariableAssignment -> Doc a
 prettyUserVariableAssignment (UserVariableAssignment assignment) = do
-  let prettyLine (var, value) = pretty var <> ":" <+> pretty value
   vsep (fmap prettyLine assignment)
+  where
+    prettyLine a = do
+      case a of 
+        TensorAssignment (var, value) -> pretty var <> ":" <+> pretty value
+        RecordAssignment (var, fields) -> pretty var <> ":" <+> "{" <+> fillSep (fmap (\(name, tens) -> pretty name <+> "=" <+> pretty tens) fields) <+> "}"
```
MatthewDaggitt: Can we use the `Pretty` instance we defined for `UserVariableAssignment` instead of reimplementing it here?

vehicle/src/Vehicle/Backend/Solver/QueryCompilation/InputBoundsCheck.hs
```diff
@@ -202,6 +202,7 @@ checkAllBoundsPresent (Partial allPartialbounds assertions) = do
     [] -> return ()
     i : is -> do
       nameCtx <- getCompleteNamedCtx
+      _ <- logDebug MaxDetail $ pretty (show nameCtx)
```
MatthewDaggitt: Remove.

vehicle/src/Vehicle/Backend/Solver/UserVariableElimination/VariableReconstruction.hs
```diff
@@ -37,15 +38,48 @@ reconstructUserVars ::
...
+  (MonadLogger m) =>
+  UserVariableAssignment ->
+  [CompilationStep] ->
+  m UserVariableAssignment
+reconstructRecords existingAssignment steps = do
+  foldlM checkStep existingAssignment steps
+  where
+    checkStep (UserVariableAssignment assignments) step = do
+      case step of
+        ConvertQuantifiedTensorLike tensorName recordName fieldNames -> do
```
MatthewDaggitt, line 62: This function should disappear and the code from below this point is the code that should be called in the `ConvertQuantifiedTensorLike` in the `applyReconstructionStep` function below?

vehicle/src/Vehicle/Backend/Solver/UserVariableElimination/VariableReconstruction.hs
```diff
@@ -37,15 +38,48 @@ reconstructUserVars ::
...
+                    _ -> False)
+                  assignments
+
+          tensorValue <- case tensorAssignment of
+            Just (TensorAssignment ( _ , t)) -> return t
+            -- TODO: proper/better error message
+            _ -> developerError "No assignment found"
+
+          let fieldIndices = [0 .. length fieldNames - 1] :: [Int]
+          let tensorIndices = map (\i -> at tensorValue i) fieldIndices
```
MatthewDaggitt: I think you just want to call `unstack` from `Vehicle.Data.Tensor` which split the tensor into a list of the sub-tensors along the first dimension. Then you can just zip them with the names.

vehicle/src/Vehicle/Backend/Solver/UserVariableElimination/VariableReconstruction.hs
```diff
@@ -37,15 +38,48 @@ reconstructUserVars ::
...
+  foldlM checkStep existingAssignment steps
+  where
+    checkStep (UserVariableAssignment assignments) step = do
+      case step of
+        ConvertQuantifiedTensorLike tensorName recordName fieldNames -> do
+          let tensorAssignment = find
+                  (\case
+                    TensorAssignment (tn, _) -> tn == tensorName
+                    _ -> False)
+                  assignments
```
MatthewDaggitt: Should be able to use `Map.lookup` on `assignments` rather than do a linear scan through the assignments (which should be a `MixedVariableAssignments` object once you move it to the code below).

vehicle/src/Vehicle/Compile/Print.hs
```diff
@@ -616,6 +616,9 @@ instance
       prettyUsing @restVar (toSliceVar var, ctx)
         <+> "->"
         <+> pretty d
+        -- TODO: RECORD SUPPORT
```
MatthewDaggitt: Remove TODO?

vehicle/tests/golden/features/tensorLike-quantifier/test.json
```diff
@@ -17,5 +17,11 @@
     "lines": ".*\"fileHash\".*"
   },
   "sizeOnly": ".vcl-plan"
-}
+  },
+  {
+    "name": "MarabouVerify",
+    "run": "vehicle verify -s spec.vcl -v TestVerifier --network f:fake.onnx --property p --verifier-location=testVerifier --verifier-args='sat 'f_X[0]=0.2,f_X[1]=0.2,f_Y[0]=0.3,f_Y[1]=3.0'' --logging MaxDetail",
```
MatthewDaggitt: Remove `--logging MaxDetail`?

vehicle/tests/golden/features/tensorLike-quantifier/spec.vcl
```diff
@@ -13,10 +13,35 @@ maxBound = { a = 10, b = 10 }
 @network
 f : Pair -> Pair
 
+-- @property
+-- p : Bool
+-- p = forall x . (minBound + minBound) <= x <= maxBound => (f x).a > x.a
```
MatthewDaggitt: Remove all commented out code if not using?

vehicle/src/Vehicle/Verify/Specification/Execute.hs
```diff
@@ -324,10 +323,21 @@ writeWitnessToFile :: (MonadVerify m) => FilePath -> PropertyAddress -> UserVari
...
+      handleAssignment a folder = do
+        case a of
+          TensorAssignment (var, tensor) -> do
+            let file = folder </> layoutAsString (pretty var)
+            let dims = Vector.fromList (shapeOf tensor)
+            -- TODO got to be a better way to do this conversion...
+            let unboxedVector = Vector.fromList $ BoxedVector.toList (fmap realToFrac (Tensor.toVector tensor))
+            let idxData = IDXDoubles IDXDouble dims unboxedVector
+            liftIO $ encodeIDXFile idxData file
+            -- TODO: IDX is a file format for vectors and matrices - i think we want to convert back to a tensor here?
```
MatthewDaggitt: Yup this makes sense. You could split these two cases out into two sub-functions in the `where` block?

vehicle/src/Vehicle/Verify/Core.hs
```diff
@@ -176,13 +176,28 @@ flipQueryRel = \case
...
 
 instance FromJSON UserVariableAssignment
 
-instance Pretty UserVariableAssignment where
-  pretty (UserVariableAssignment assignment) =
-    vsep (fmap pretty assignment)
+data UserVariableAssignmentType 
+  = TensorAssignment (Name, RatTensor)
+  -- Attempting to cheat recordfields here
+  | RecordAssignment (Name, [(Name, RatTensor)])
```
MatthewDaggitt: Probably want to make this a `NonEmpty` list rather than a standard list. Will stop you from having to use the partial function `head` later on.

vehicle/src/Vehicle/Compile/Print/Error.hs
```diff
@@ -918,24 +918,49 @@ formatCompileError = \case
...
-              <+> "for"
-              <+> squotes (prettyFriendly (WithContext inputValue ctx)),
-        fix =
-          Just $
-            "add additional inequalities that restrict the value of" <+> case userVariables of
-              [v] -> "the quantified variable" <+> quotePretty v
-              _ -> "the following quantified variables:" <+> hsep (fmap pretty userVariables)
-      }
+    case toTypeValue inputValue of 
+      (VRecordType _t fields) -> do
```
MatthewDaggitt: Could we do the case split a bit lower down? A lot of the text here is shared. It would make sense to do the case splits in a `where` block below for the specific subcomponents that differ?

vehicle/src/Vehicle/Backend/Solver/UserVariableElimination.hs
```diff
@@ -71,13 +74,18 @@ eliminateExists (QuantifyRatTensorArgs _ binder (Closure env body)) = do
     -- Recursively compile the expression.
     (partitions, networkInputEqualities) <-
       logCompilerSection2 MidDetail "reduction of body to assertion tree" $ runWriterT (compileBoolExpr normExpr)
+    
+    partitions' <- case partitions of
+      Trivial b -> pure (Trivial b)
+      NonTrivial (Partitions m) ->
+        pure (NonTrivial (Partitions (Map.mapKeys (prevSteps ++) m)))
```
MatthewDaggitt: Can we abstract this code into a function of type `CompilationStep -> MaybeTrivial Partitions -> MaybeTrivial Partitions` and call it when we pattern match on the record quantifiers rather than passing the compilation steps through to `eliminateExists`? 

You can also use `traverse _ _` instead of pattern matching on `Trivial`/`NonTrivial`.

vehicle/src/Vehicle/Backend/Solver.hs
```diff
@@ -289,7 +296,12 @@ wrapQuantifyRecord QuantifyRecordArgs {..} = do
 
   -- Construct body (_PairFromTensor _t0)
   let nestedBody = App recordQLam [Arg Explicit Relevant fromTensorExpr]
-  return $ QuantifyRatTensorArgs normalisedDims tensorBinder (Closure boundEnv nestedBody)
+  let ratTensorArgs = QuantifyRatTensorArgs normalisedDims tensorBinder (Closure boundEnv nestedBody)
+  
+  fieldNames <- getRecordFieldNames recordTypeIdent
+  let name = fromMaybe (developerError "Quantified variable binder should have name") (nameOf quantifyRecordBinder)
```
MatthewDaggitt: You can use `getBinderName` (which cheats in the same way)

MatthewDaggitt (changes requested):
Generally looks great and definitely the right approach! I've added a bunch of minor representation changes and cleanups that will improve it, but nothing major I hope.

MatthewDaggitt (approved):

PR #1153, opened 2026-06-01
Update vehicle-rocq for compatibility

lstrsrmn:
Dune was having issues compiling `utils.v`, so I rearranged the `vehicle-rocq` package so that dune does not try to build `uilts.v`, and instead `make` builds it, avoiding the issue.


MatthewDaggitt:
@lstrsrmn this definitely works now? Ready for merging?

MatthewDaggitt (approved):

lstrsrmn:
It has worked for my computer and a VM, but I would like @hoheinzollern to confirm it works for them first.

MatthewDaggitt:
Merging in so we can make a release

PR #1156, opened 2026-06-03
fix: package def and build instructions for Rocq backend

hoheinzollern:
The Rocq backend currently depends on the master branch of upstream mathcomp / analysis / finmap.
This PR pins the Rocq backend to the specific hashes of those packages that are known to work properly.
If not merged before a mathcomp release, it is intended to be updated to the released versions of mathcomp & co.
Enables caching to save ~45 mintues of build time per platform.


hoheinzollern:
Turned this into a draft PR to match the discussions in https://github.com/math-comp/math-comp/pull/1610 and https://github.com/math-comp/math-comp/pull/1606

hoheinzollern:
CI green, ready to merge @MatthewDaggitt

PR #1165, opened 2026-06-14
Fixed implementation of `Vector` operations for loss backend

MatthewDaggitt:
Still not quite convinced it is giving the right answer, but at least the `mnist-robustness` specification compiles and the loss function now runs.


MatthewDaggitt:
Okay so the bugs are appearing in the translation of index comparisons. Fixing that should happen in a second PR so for the moment, I have just disabled the index comparisons.

jh4n23:
Has there been any progress with index comparisons? I'm also trying to generate a loss function from the MNIST robustness property and getting stuck here

MatthewDaggitt:
Created an issue for this (https://github.com/vehicle-lang/vehicle/issues/1183) with an update. Feel free to post further questions or comments there.

PR #1170, opened 2026-06-16
Feat/tensor like loss

ggustavs:
Adds the ability to use tensor records with the loss backend.
Static typing with frozen dataclasses generated from the vcl spec is supported via a new vehicle CLI command `vehicle compile python-types`, which outputs a new `.py` file.
The record types can then be used by importing them from the generated file.
See [this example](https://github.com/vehicle-lang/vehicle/blob/feat/tensorLike-loss/examples/windController-newStyle/train.py) for the intended workflow.


wenkokke:
@ggustavs Shouldn't this be generating a `.pyi` file?

PR #1175, opened 2026-06-20
fix(rocq): align notation precedences/associativity with mathcomp

### LINKED ISSUE #1174: Operator precedence mismatch in the Rocq backend
Agent-tested operators results in the following:

| Construct (Rocq.hs) | Generated | Annotated (lvl, assoc) | Actual Coq (lvl, assoc) | Status |
|---|---|---|---|---|
| `And` | `_ && _` | 40, Left | 40, Left | ✅ correct |
| `Or` | `_ \|\| _` | 50, Left | 50, Left | ✅ correct |
| `Implies` | `_ ==> _` | 55, Right | 55, Right | ✅ correct |
| `Cons` | `_ :: _` | 60, Right | 60, Right | ✅ correct |
| `Add` / `Sub` | `_ + _`, `_ - _` | 50, Left | 50, Left | ✅ correct |
| `Mul` / `Div` | `_ * _`, `_ / _` | 40, Left | 40, Left | ✅ correct |
| comparisons | `_ <= _`, `_ < _`, `_ == _`, … | 70, NotAssoc | 70, no-assoc | ✅ correct |
| `PropNot` | `~ _` | 75, Right | 75, Right | ✅ correct |
| `PropAnd` | `_ /\ _` | 80, Right | 80, Right | ✅ correct |
| `PropOr` | `_ \/ _` | 85, Right | 85, Right | ✅ correct |
| `Not` (negb) | `~~ _` | 35, **Left** | 35, **Right** | ⚠️ assoc wrong |
| `AtTensor` | `_ ^^ _` | 30, **Left** | 30, **no-assoc** | ⚠️ assoc wrong |
| `Neg` (opp) | `- _` | **80**, NotAssoc | **35**, Right | ❌ level wrong |
| `If` | `if _ then _ else _` | **0**, NotAssoc | **200** | ❌ level wrong |
| `compileFunctionType` | `_ -> _` | **100**, Right | **99**, Right | ❌ off-by-one |
| quantifiers / `fun` | `forall _, _`, `fun _ => _` | **100** | **200** | ❌ level wrong |
| `compileTensorLiteral` | `const_t …`, `ntensor_of_tuple …` | **Nothing** | application (0) | ❌ never brackets as arg |

I'm preparing a PR to fix the mismatch and drop extra parentheses.

hoheinzollern:
The Rocq backend annotated several notations with precedences that don't match the real Coq/mathcomp levels, and `compileNotationAndArgs` already received an `Associativity` argument (passed by all call sites) but discarded it (`_associativity`), so `bracketArgs` was associativity-blind. Because the bracketing test is conservative (parenthesise on `>=`), most of the resulting errors were redundant parentheses, but two wrong levels were latent mis-parses that current specs happened not to trigger.

Corrected levels/associativity (truth via `Print Notation`, mathcomp master):
- if/then/else:       0  -> 200   (was under-bracketed as an operand)
- unary minus (opp):  80 -> 35, NotAssoc -> Right
- forall / fun:       100 -> 200
- (->):               100 -> 99
- tensor literals:    Nothing -> application precedence (bracket as an argument)
- negb (~~):          Left -> Right
- nindex (^^):        Left -> NotAssoc

Validation: all Rocq compile goldens regenerated; full RocqVerify suite compiles the de-bracketed output against the pinned mathcomp. Rocq.hs-only change.

fixes: #1174

### DIFF
--- a/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
+++ b/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
@@ -223,7 +223,7 @@ compileApplication dependencies fun args = do
       then return (getPrecedence fun, fun)
       else do
         compiledArgs <- traverseArgs compileExpr args
-        bracketedArgs <- bracketArgs functionApplicationPrecedence compiledArgs
+        bracketedArgs <- bracketArgs (replicate (length compiledArgs) functionApplicationPrecedence) compiledArgs
         return (functionApplicationPrecedence, hsep (fun : bracketedArgs))
 
   return $ annotate (Set.fromList dependencies, precedence) annDoc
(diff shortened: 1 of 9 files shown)


MatthewDaggitt (approved):
Seems reasonable and the tests pass.

PR #1181, opened 2026-06-23
Try to fix Python CI

MatthewDaggitt:


MatthewDaggitt:
Okay so this problem only occurs on `musllinux` and not `manylinux`...

MatthewDaggitt:
The problem appears to be this line:
```
ValueError: setuptools-scm was unable to detect version for
/tmp/pip-install-0hlnmkd1/ninja_65d7c31f53cb44e89496b1ce9e879415
```
No new release of ninja seems to be have been made recently so unsure what has changed upstream. I have tried pinning to an earlier version of ninja without success.

MatthewDaggitt:
Okay, I appear to have found a fix that works by pre-installing `ninja` before we build the wheel but it is definitely not a principled one... 

@wenkokke would appreciate your thoughts on this one?

wenkokke:
> Okay, I appear to have found a fix that works by pre-installing `ninja` before we build the wheel but it is definitely not a principled one... 
> 
> 
> 
> @wenkokke would appreciate your thoughts on this one?

The principled approach is probably me rebuilding the docker image.

wenkokke:
@MatthewDaggitt I've hit the button to regenerate the image in question, but you might wanna try using any of the following containers instead, especially if you're considering moving the recommended GHC version or minimum musllinux version:

- https://hub.docker.com/r/wenkokke/musllinux_1_1_ghc967_x86_64
- https://hub.docker.com/r/wenkokke/musllinux_1_1_ghc984_x86_64
- https://hub.docker.com/r/wenkokke/musllinux_1_2_ghc965_x86_64
- https://hub.docker.com/r/wenkokke/musllinux_1_2_ghc967_x86_64
- https://hub.docker.com/r/wenkokke/musllinux_1_2_ghc984_x86_64

If you're changing the recommended GHC version, e.g., to GHC 9.6.7, you should probably change _all_ the images and CI instances to use it, though. Best to be consistent. You can see all the images that I've made available here:

- https://hub.docker.com/u/wenkokke

MatthewDaggitt:
Thanks @wenkokke!

I wasn't planning to bump the default GHC version, but will keep that in mind if we do.

wenkokke:
> Thanks @wenkokke!
> 
> 
> 
> I wasn't planning to bump the default GHC version, but will keep that in mind if we do.

Just to be sure, please retry the unmodified branch, as I've rebuilt the image.

MatthewDaggitt:
Set all jobs to rerun.

MatthewDaggitt:
Rerun jobs have passed so closing

wenkokke:
> Rerun jobs have passed so closing

Yay I fixed it

PR #1185, opened 2026-06-26
Prepare to Compile Multimodal Network Shapes

DanielEJones:
This PR introduces the idea of network modalities to the compiler and stubs out all the places that will potentially need changes to support this.


MatthewDaggitt (approved):
Looks good to me!

PR #1188, opened 2026-07-01
WIP CompareRatTensor: Refactoring of Pointwise & Reduced Tensor Comparisons

jaackies:
Working on refactoring `CompareRatTensorPointwise` and `CompareRatTensorReduced` into the same representation `CompareRatTensor` - _still WIP!_

Changes:
- Removed builtin functions `CompareRatTensorPointwise` and `CompareRatTensorReduced`, replaced with`CompareRatTensorPointwise` in Core, Interface, Type, Args, Accessors.
- `append` function added back as a Vehicle function for use with the `CompareRatTensor` Type
- Pushed `CompareRatTensor` into Desugaring, Resugaring, Interface/Normalise and TypedView
  - _`evalCompareRatTensor` in Normalise is still WIP_

As a note, there are some large commits from when benchmarking files from a different branch somehow made it onto this branch (necessitating their removal later on).


vehicle/src/Vehicle/Compile/Sugar/Desugar.hs
```diff
@@ -420,12 +401,13 @@ elabExpr expr = case expr of
   B.Lt e1 tk e2 -> standardLibComparison V.Lt tk e1 e2
   B.Ge e1 tk e2 -> standardLibComparison V.Ge tk e1 e2
   B.Gt e1 tk e2 -> standardLibComparison V.Gt tk e1 e2
-  B.EqPoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Eq) tk [e1, e2]
-  B.NePoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Ne) tk [e1, e2]
-  B.LePoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Le) tk [e1, e2]
-  B.LtPoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Lt) tk [e1, e2]
-  B.GePoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Ge) tk [e1, e2]
-  B.GtPoint e1 tk e2 -> builtinFunction (V.CompareRatTensorPointwise V.Gt) tk [e1, e2]
+  -- need to work out best way to allow for +e1, e2 into the list from function
```
MatthewDaggitt: These needs to be fixed! You can pass through `e1 e2` as a list as an extra argument to the function and then append them.

vehicle/lib/Definitions.vcl
```diff
@@ -19,40 +19,56 @@ forallInList f xs = fold (\x y -> x and y) True (map f xs)
 existsInList : (A -> Bool) -> List A -> Bool
 existsInList f xs = fold (\x y -> x or y) False (map f xs)
 
+-------
+
+append : List A -> List A -> List A
+append xs ys = fold (\y x -> x:y) ys xs
```
MatthewDaggitt: `:` should be written `::` in Vehicle.

vehicle/lib/Definitions.vcl
```diff
@@ -19,40 +19,56 @@ forallInList f xs = fold (\x y -> x and y) True (map f xs)
...
+-- append xs ys = fold (\x y -> x++y) xs ys -- alt, appending at end
```
MatthewDaggitt: Remove the comment?

vehicle/lib/Definitions.vcl
```diff
@@ -19,40 +19,56 @@ forallInList f xs = fold (\x y -> x and y) True (map f xs)
...
 --------------------------------------------------------------------------------
 -- Tensor
 --------------------------------------------------------------------------------
 -- These operations have non-zero dimensions so that we have a unique
 -- representation of relationships between zero-dimensional tensors
 -- (i.e. pointwise comparison).
 
-eqRatTensorReduced : Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
-eqRatTensorReduced xs ys = reduceAnd (xs ==. ys)
+eqRatTensorReduced : {dim: Nat} -> {dims : List Nat} -> Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
```
MatthewDaggitt: No need to add these implicit arguments. They will be automatically inserted by the compiler.

vehicle/lib/Definitions.vcl
```diff
@@ -19,40 +19,56 @@ forallInList f xs = fold (\x y -> x and y) True (map f xs)
...
 
 geRatTensorReduced : Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
-geRatTensorReduced xs ys = reduceAnd (xs >=. ys)
+geRatTensorReduced xs ys = reduceAnd True (xs >=. ys)
 
 gtRatTensorReduced : Tensor Real (dim :: dims) -> Tensor Real (dim :: dims) -> Bool
-gtRatTensorReduced xs ys = reduceAnd (xs >. ys)
+gtRatTensorReduced xs ys = reduceAnd True (xs >. ys)
+
+--------------------------------------------------------------------------------
```
MatthewDaggitt: I think all the changes in this file below this point have been pulled in from a different commit somehow and should be removed!

vehicle/src/Vehicle/Compile/Sugar/Desugar.hs
```diff
@@ -568,6 +553,29 @@ elabTypeBinder elab folded = \case
   B.InstanceTypeBinder modalities t -> mkBinder elab folded modalities (V.Instance False) $ That t
   B.BasicTypeBinder b -> elabBasicBinder elab folded b
 
+-- re: elabCompareReduced and elabComparePointwise, even though Pointwise and Reduced Tensor Comparisons are distinct in the frontend/user syntax,
+-- the compiler moves them to a single representation, CompareRatTensor that is able to handle both pointwise and reduced.
+
+elabCompareReduced :: (MonadElab m) => B.Expr -> m (V.Expr Builtin)
+elabCompareReduced = \case
+  B.CompareRatTensorReducedEq tk -> builtinFunction (V.CompareRatTensor V.Eq) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
```
MatthewDaggitt: There's still a lot of duplicating in this function. Instead of redoing the case here, pass the correct operation (`V.Ne` etc.) through as an argument to this function.

vehicle/src/Vehicle/Compile/Sugar/Desugar.hs
```diff
@@ -568,6 +553,29 @@ elabTypeBinder elab folded = \case
...
+  B.CompareRatTensorReducedNe tk -> builtinFunction (V.CompareRatTensor V.Ne) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
+  B.CompareRatTensorReducedLe tk -> builtinFunction (V.CompareRatTensor V.Le) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
+  B.CompareRatTensorReducedLt tk -> builtinFunction (V.CompareRatTensor V.Lt) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
+  B.CompareRatTensorReducedGe tk -> builtinFunction (V.CompareRatTensor V.Ge) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
+  B.CompareRatTensorReducedGt tk -> builtinFunction (V.CompareRatTensor V.Gt) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims"]
+  expr -> elabExpr expr
+
+elabComparePointwise :: (MonadElab m) => B.Expr -> m (V.Expr Builtin)
+elabComparePointwise = \case
+  B.CompareRatTensorPointwiseEq tk -> builtinFunction (V.CompareRatTensor V.Eq) tk [B.Hole $ mkToken B.HoleToken "_pDims", B.Nil $ mkToken B.TokNil "nil"]
```
MatthewDaggitt: Same comment here!

MatthewDaggitt (commented):
Haven't quite finished reviewing yet, but will try to do so tomorrow. Feel free to address some of these points beforehand though!

examples/windController/agdaProof/WindControllerSpec.agda
```diff
@@ -0,0 +1,42 @@
+-- WARNING: This file was generated automatically by Vehicle
```
MatthewDaggitt: This file should be deleted.

vehicle/lib/Definitions.vcl
```diff
@@ -19,6 +19,11 @@ forallInList f xs = fold (\x y -> x and y) True (map f xs)
 existsInList : (A -> Bool) -> List A -> Bool
 existsInList f xs = fold (\x y -> x or y) False (map f xs)
 
+-------
+
+append : List A -> List A -> List A
+append xs ys = fold (\y x -> x :: y) ys xs
```
MatthewDaggitt: Should be `(\x y -> x :: y)`

vehicle/src/Vehicle/Compile/Sugar/Desugar.hs
```diff
@@ -568,6 +568,16 @@ elabTypeBinder elab folded = \case
   B.InstanceTypeBinder modalities t -> mkBinder elab folded modalities (V.Instance False) $ That t
   B.BasicTypeBinder b -> elabBasicBinder elab folded b
 
+-- re: elabCompareReduced and elabComparePointwise, even though Pointwise and Reduced Tensor Comparisons are distinct in the frontend/user syntax,
+-- the compiler moves them to a single representation, V. CompareRatTensor, that is able to handle both pointwise and reduced comparisons
+elabCompareReduced :: (MonadElab m, IsToken token) => V.ComparisonOp -> token -> [B.Expr] -> m (V.Expr Builtin)
+elabCompareReduced op tk [e1, e2] = builtinFunction (V.CompareRatTensor op) tk [B.Nil $ mkToken B.TokNil "nil", B.Hole $ mkToken B.HoleToken "_rDims", e1, e2]
```
MatthewDaggitt: Gah I've found a bug here. We need the dimensions token to be implicit... so we can't use `builtinFunction`. Instead, we need to manually expand the `builtinFunction` until we get to a point where we can provide the explicit/implict status...

MatthewDaggitt (changes requested):

vehicle/lib/Definitions.vcl
MatthewDaggitt: These 6 definitions should all be redundant! As they are just the reduce case.

vehicle/src/Vehicle/Data/Builtin/Interface/Normalise.hs
```diff
@@ -476,20 +476,39 @@ evalReduceMinRatTensor = evalReduceTensor accessReduceMinRatBuiltin accessRatTen
...
-    Nothing
-    Nothing
-    Nothing
-    Nothing
+evalCompareRatTensor :: forall builtin m. (MonadNormBuiltin m, HasBoolExpr Value builtin, BuiltinHasBoolType builtin, HasRatExpr Value builtin, PrintableBuiltin builtin) => ComparisonOp -> EvalSimple TensorComparisonArgs Value builtin m
+evalCompareRatTensor op = \case
+  -- base case where we are up to pointwise comparison (dims1/pDims = [])
+  TensorComparisonArgs IDimNil rDims xs ys ->
+    -- xs and ys passed on to evalHeteteroTensorOp2 to handle
+    evalHeteroTensorOp2 (mkExpr accessCompareRatTensorBuiltin op) accessRatTensorLiteral accessBoolTensorLiteral (comparisonOp op) Nothing Nothing Nothing Nothing (TensorOp2Args rDims xs ys)
```
MatthewDaggitt: I think I've spotted another bug here. In `evalHeteroTensorOp2` if we can't evaluate the pointwise comparison then `evalSimple` is going to reconstruct the expression with `TensorOp2Args rDims xs ys` instead of `TensorComparisonArgs IDimNil rDims xs ys` (i.e. the IDimNil is going to be incorrectly dropped). To fix this I think we need to pass both the default set of arguments through as a list of type `Spine builtin` (constructed via `mkExpr accessSpine (TensorComparisonArgs IDimNil rDims xs ys)`) and then expand out the use of `evalSimple` to use the default list instead of the args when `eval` returns `Nothing`. 

The definition of `evalTensorOp2` can then simply exactly the same arguments through (you will need to expand the list of the arguments and then call `mkExpr accessSpine args`).

MatthewDaggitt (commented):
Two more bugs

MatthewDaggitt (approved):

MatthewDaggitt:
Okay, I think I've got all the remaining bugs in the other parts of the compiler not on the solver path.

PR #1190, opened 2026-07-03
Fix: record quantification in Rocq backend for Issue#1189

misr01:
Fixes the error that comes up when trying to export the [windControllerNew ](https://github.com/vehicle-lang/vehicle/tree/dev/examples/windController-newStyle) to Rocq:

```
 Error: Quantification over TensorLikes is unsupported.
Stack:
CallStack (from HasCallStack):
  developerError, called at src/Vehicle/Compile/Error.hs:283:35 in vehicle-0.26.0-inplace:Vehicle.Compile.Error
  unsupportedTensorLikeQuantifier 
```

- This was done by reusing the `QuantifyRatTensor` binder logic in `QuantifyRecord.`
- A new helper function `compileQuantifierFunction` was created to make the code cleaner and keep it more consistent with the [Agda](https://github.com/vehicle-lang/vehicle/blob/dev/vehicle/src/Vehicle/Backend/ITP/Agda.hs) backend.
- Removed unsupportedArgsError from compileBuiltin's where block since the new helper handles errors and it is not used anywhere else.

Tested the fix locally on a few small specifications and the windControllerNew example, and it successfully generates working Rocq proofs.

### DIFF
--- a/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
+++ b/vehicle/src/Vehicle/Backend/ITP/Rocq.hs
@@ -552,16 +552,14 @@ compileBuiltin b args = case b of
     ReduceMaxRatTensor -> unsupportedError
     ReduceMulRatTensor -> compileApplication [] "reduceMul" args
     ConstTensor -> compileApplication [MathcompImport Algebra] "const_t" args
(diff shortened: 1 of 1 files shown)


MatthewDaggitt:
Thanks, this looks great! Merging in.

PR #1200, opened 2026-07-10
update backend support badges for parameters.rst

jh4n23:


MatthewDaggitt:
Thanks, looks great! Is there any chance you could add a couple of sentences to the training documentation that describe the limitations that you currently have to pass it explicitly, and maybe add an arrow link next to the badge that links to that section like we do for the limitations on the multiplication operations?

I've also invited you as a collaborator so you don't have to work from a fork of the repo.

docs/training.rst
```diff
@@ -77,6 +77,7 @@ TensorFlow works the same way—load the declarations through ``vehicle_lang.los
 
 By combining your usual task loss with the constraint losses returned by Vehicle, you ensure the optimiser simultaneously keeps the model on task and enforces the specification.
 
+Note that calling ``constraint_loss_fn()`` will require passing the relevant parameters, datasets, and networks that are mentioned in the specification. This may include inferred parameters. For example, the size of a dataset can be inferred by the specification at compile time, and therefore may be tagged with ``infer=True``. However, since the dataset is passed to the loss backend at runtime, this value cannot be inferred. Therefore, it must still be passed explicitly as a parameter.
```
MatthewDaggitt: Can we take out the "may"s? And change it to "this value cannot currently be inferred"

MatthewDaggitt (commented):
Otherwise looks great!

docs/training.rst
jh4n23: Sure - I don't know about an example case where two properties are specified, but only one makes use of an inferred parameter, and the loss function was being generated using the other property. Would the inferred parameter still need passing, or would it never raise an issue as its value is never checked?

I did spend a couple of minutes trying to test this, but encountered too many other errors along the way and decided to save time by writing ambiguously just in case it made a difference.

MatthewDaggitt: The unused inferred parameter shouldn't need passing in that case.

PR #1202, opened 2026-07-13
Generate Record Comparison Instances

DanielEJones:
This PR adds support for basic comparison operations (`<`, `<=`, `>`, `>=`, `==`, `!=`) between two records of the same type.


vehicle/src/Vehicle/Compile/Scope/Records.hs
```diff
@@ -119,6 +122,9 @@ createRecordProjectionFn p ident telescope visibility (field, fieldType) = do
 -- ValidNetworkIOType instance generation
 --------------------------------------------------------------------------------
 
+instanceBinder :: Provenance -> Name -> Expr Builtin -> GenericBinder (Expr Builtin)
+instanceBinder p name = Binder (BinderDisplayForm (NameAndType name p) True) (Instance True) Relevant
```
MatthewDaggitt: Could this be added to the `Binder` file? If there isn't already such a function there...

vehicle/src/Vehicle/Compile/Scope/Records.hs
```diff
@@ -162,16 +168,109 @@ createRecordHasValidIOTypeInstance p recordIdent telescope fields = do
...
+  --      { f1 : t1
+  --      , f2 : t2
+  --      }
+  --
+  -- We need to generate methods of the form:
+  --
+  --    @instance
+  --    recordPairHasComparisonInstance :
+  --      T1 -> T2 ->
+  --      HasComparisonInstance T1 T1 -> HasComparisonInstance T2 T2 ->
```
MatthewDaggitt: Very minor but could we put each argument on a separate line?

MatthewDaggitt (changes requested):
Otherwise looks great!

MatthewDaggitt (approved):

Issue #1196, opened 2026-07-07
Clarify vehicle export vs vehicle compile itp in ITP export documentation


misr01:
Hi, I noticed a potential documentation mismatch around the "Exporting to ITPs" section.

The "Exporting a specification to an ITP" documentation explains the `vehicle --export` command, but the current CLI also provides vehicle compile itp. I could not find `vehicle --compile itp` documented as the recommended command, or an explanation of how it differs from vehicle export.

Vehicle export requires a --cache argument, while vehicle compile itp allows a cache to be provided optionally. Could the documentation clarify whether:

- vehicle export is still the preferred command, or
- vehicle compile itp is the intended/new command for exporting to ITPs?

MatthewDaggitt:
Nice catch! The intended command is `vehicle export ...`. Would you be able to open a PR updating the documentation?

(The `vehicle compile itp` mode is supposed to be for internal use only and is used to implement `vehicle export` and therefore shouldn't be in the documentation).

lstrsrmn:
@MatthewDaggitt I think there is a use for `vehicle compile itp`. It is essentially just `vehicle export` but it requires no `verify` step, and thus no network. I think this would be especially useful if the data/neural network is under protection, as it would allow formal methods engineers to still produce and verify integration of the vehicle specification with a greater system proof. Is the intention to make `vehicle export` have another mode which acts almost exactly as `vehicle compile itp` does, or something similar?

MatthewDaggitt:
So the idea is that `vehicle compile` is supposed to be entirely an internal command that isn't seen by the user.

You're right, it is useful to be able to generate the specification without referencing the cache, so we should make `cache` option to `vehicle export` optional when we do this (change the type to `Maybe FilePath`).
https://github.com/vehicle-lang/vehicle/blob/dbbbbd65eb05b73f738bc022f63e7d9ba7ed47c3/vehicle/src/Vehicle/Export.hs#L14

It would also be be great to Rocq command at https://github.com/vehicle-lang/vehicle/tree/dev/examples/windController

PR #1228, opened 2026-08-19
Support specification/cache exclusivity for vehicle export

misr01:
implements the `vehicle export` CLI behaviour discussed in #1196 

- require exactly one of `--cache` or `--specification`
- allow `--declaration`, `--network`, `--dataset`, and `--parameter` when
  exporting directly from a specification
- preserve the existing behaviour when exporting from a verification cache
- reject specification-specific resource arguments when `--cache` is used
- provide specific errors when both inputs or neither input are supplied
- add command-line parser tests for both export modes

Related to #1196

### DIFF
--- a/vehicle/src/Vehicle/CommandLine.hs
+++ b/vehicle/src/Vehicle/CommandLine.hs
@@ -353,9 +353,14 @@ exportParser :: Parser ExportOptions
 exportParser =
   ExportOptions
     <$> itpParser
-    <*> exportCacheParser
+    <*> optional specificationParser
+    <*> declarationParser
+    <*> networkParser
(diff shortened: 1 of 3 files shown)


MatthewDaggitt (changes requested):
Thank you this looks great! Three minor pieces of feedback:

1. We need to add an entry to the CHANGELOG.md under `## Next release` describing the change.
2. The test is failing so you need to accept the new printed value (see the `CONTRIBUTION.md` file on how to update the golden tests.
3. We need to update `vehicle_lang/export/__init__.py` to have the same interface (and add an entry in the CHANGELOG).

PR #1241, opened 2026-08-31
Add Python tests to CI

MatthewDaggitt:
For some reason the Python tests are never actually on run on CI. This will keep me honest and hopefully avoid future breakages.

### DIFF
--- a/.github/workflows/build-vehicle-python.yml
+++ b/.github/workflows/build-vehicle-python.yml
@@ -71,7 +71,7 @@ jobs:
           CIBW_BUILD: "*${{ matrix.os.plat }}*"
           CIBW_ARCHS: "${{ matrix.os.arch }}"
           CIBW_TEST_EXTRAS: "${{ matrix.os.test }}"
-          CIBW_TEST_COMMAND: "python -m pytest {project}/tests"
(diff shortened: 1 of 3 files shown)


MatthewDaggitt:
@wenkokke it looks the `pty` library is not available on Windows. I'm not sure I feel confident enough to rewrite the `Session.check_output` method that you wrote to avoid using it. Do you have any thoughts or advice on it? Should it be theoretically possible to do using OS agnostic libraries?

wenkokke:
> @wenkokke it looks the `pty` library is not available on Windows. I'm not sure I feel confident enough to rewrite the `Session.check_output` method that you wrote to avoid using it. Do you have any thoughts or advice on it? Should it be theoretically possible to do using OS agnostic libraries?

Wasn't me. That one was @ggustavs. So I think it's definitely possible to add Windows support back in.

MatthewDaggitt:
Ah sorry I just assumed, apologies. Doing the proper archaeology it looks like it is this commit that is to blame: https://github.com/vehicle-lang/vehicle/commit/b4fa78539330230e0c0e2fa99fbb5d01064d6cab#diff-7f67a804772cef203fb6fd30b0c7c908917aec787145f220882e07e41f0a0d9d

PR #1254, opened 2026-09-07
Remove unneeded imports in generated loss function code

### LINKED ISSUE #1193: Remove unneeded imports in generated loss function code
Currently in `translate_Main` we are adding imports such as `functools` and `fractions`. I think the former is definitely unused, and the latter may be?

zillybop:
`fractions` is still required:
```
constraint_loss_fn(network)
NameError: name 'fractions' is not defined
tests/data/test_trainable.vcl:5
```


MatthewDaggitt (approved):

MatthewDaggitt:
Thanks, looks good to me!

PR #1266, opened 2026-09-13
Fix #342 - another attempt at locking standard library and interface files properly

### LINKED ISSUE #342: Race condition in interface file loading

MatthewDaggitt:
Alternative approach to #379. Let's see if it passes the tests online when running in a multi-threaded environment.


MatthewDaggitt:
Looks like it's working!

PR #1268, opened 2026-09-14
Checking for substring "Status: verified" instead of just "verified" in the Agda backend

rakshikab:
Since "verified" is a sub string of both "Status: verified" as well as "Status: unverified", both were accepted in the Agda backend. Updating the sub string check to fix this.


MatthewDaggitt (approved):

MatthewDaggitt:
Thanks a lot! Really appreciate the fix!

PR #1271, opened 2026-09-16
Fix #1251 - Validating empty domains before calling samplers in loss backend

### LINKED ISSUE #1251: Samplers don't correctly handle the case where the domain is empty
Currently samplers don't check that the domains are actually non-empty and if a domain is calculated as `(1, -1)` for example, then it will generate incorrect samples.

We should really check all domains are valid once before calling the sampler, rather than relying on each sampler implementation to implement their own checks.

This can't be checked on the Haskell side as the bounds are dynamically calculated.

kkhtun:
This PR attempts a fix for #1251 
- Adds validation for sampling domains in  before invoking pytorch or tensorflow samplers.
- Rejects empty domains where a lower bound exceeds the upper bound.
- The core validation step is shared through ```_common.py```, but the actual check is backend-specific as pytorch and tensorflow represent tensor comparisons differently.

Changes to tests
- Adds regression tests in ```test_sampler_domains.py``` to ensure invalid domains are rejected before sampler execution.
- Adds an accompanying spec and golden file, covering the dynamically-bound case (using parameters).


MatthewDaggitt (changes requested):
Thank you for PR!

A couple of comments:
-  Rather than fiddling about with changing the `get_loss` functions we pass through, I think it would be conceptually neater if the `_validate_domain` methods lived on the abstract `PyTorchSampler` and `TensorFlowSampler` base clases? We could then make the existing `get_loss` method private and replace it with a public `get_loss` method that validates the domain and then calls the private one?
- Can you add a quick CHANGELOG entry describing the bug fixed?

kkhtun:
Hi @MatthewDaggitt , thanks for the feedback. I have updated my implementation to incorporate validation inside the abstract backend samplers. I have renamed the existing `get_loss()` as `_get_loss()` to act as an internal method, while a new public method `get_loss()` performs the validation and delegates to `_get_loss()`. 

One thing I noticed with this change is that this might make the existing custom samplers breaking where existing subclasses that implement `get_loss()` now need to implement `_get_loss()` instead.

Would it be preferable to keep the existing name `get_loss()` for concrete/custom sampler implementations and instead use a separate name for the public wrapper that performs the validation? Since custom samplers will need to be passed as objects without exposing `get_loss` directly (as in #1276), this could be handled internally.

PR #1276, opened 2026-09-18
Fix docs - Custom sampler example

kkhtun:
**Issue**

Previously in docs, https://vehicle-lang.readthedocs.io/en/stable/training.html#custom-samplers-and-declaration-context, we have an example `custom = {"images": MySampler().get_loss}`
which throws an attribute error since it expects the sampler object itself.

**Fix**

Update docs to
`custom = {"images": MySampler()}`
reflect the actual use case.


MatthewDaggitt:
Thanks, great spot!

MatthewDaggitt (approved):

PR #1278, opened 2026-09-20
fix: python errors through contextamanager

### LINKED ISSUE #1274: Errors raised inside a context manager are replaced by FrozenInstanceError
`VehicleInternalError` and `VehicleUserError` are declared as `@dataclass(frozen=True)`. They can be raised normally. But when one propagates out of a `@contextlib.contextmanager` block, CPython assigns `exc.__traceback__` in `_GeneratorContextManager.__exit__`, and the frozen dataclass refuses the assignment, so the original error is lost.

```python
from contextlib import contextmanager
from vehicle_lang.error import VehicleInternalError

@contextmanager
def ctx():
    yield

with ctx():
    raise VehicleInternalError("the real error")
```

gives:

```text
  File "repro.py", line 9, in <module>
    with ctx():
  File "/usr/lib/python3.11/contextlib.py", line 191, in __exit__
    exc.__traceback__ = traceback
    ^^^^^^^^^^^^^^^^^
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field '__traceback__'
```

This is reached inside `vehicle_lang` itself. `Session.check_output_pty` calls the compiler inside the `temporary_files` context manager, so when a command fails there the caller sees `FrozenInstanceError: cannot assign to field '__traceback__'` instead of the error Vehicle actually raised.

Dropping `frozen=True` on the two classes in `src/vehicle_lang/error.py` is enough to fix it. The same applies to `VehicleSessionClosed` and `VehicleSessionUsed` in `src/vehicle_lang/session/error.py`.

This behavior is reproducible with both the current PyPI release, and the dev HEAD.

ggustavs:
Fixes #1274  `contextlib` assigns to `__traceback__` when re-raising, which a frozen dataclass rejects. `eq=False` rather than a bare `@dataclass` keeps the identity equality and hashing that exceptions inherit.


MatthewDaggitt (changes requested):
Looks great to me! Could you add a small test to prevent regressions?

vehicle-python/tests/test_errors.py
```diff
@@ -0,0 +1,31 @@
...
+@contextmanager
+def context() -> Iterator[None]:
+    yield
+
+
+@pytest.mark.parametrize("error", ERRORS)  # type: ignore[untyped-decorator]
+def test_error_survives_context_manager(error: Exception) -> None:
+    with pytest.raises(type(error)):
+        with context():
+            raise error
```
MatthewDaggitt, line 33: Hmm I was hoping for a slightly more realistic test that actually involves generating an error via the Vehicle compiler/bindings that wasn't succeeding before. Is there such a test case?

ggustavs: Added one. The compiler's own errors can't reproduce this: `execute_command` raises them after `check_output` has returned, so the manager has already exited. The only path that hits the bug is an error raised inside `check_output`/`check_output_pty`, which run `check_call` inside `temporary_files`. A closed session is the reachable case.

`test_session_error_survives_temporary_files` opens a session, closes it, and calls `check_output`. In a subprocess, since closing frees the RTS for the rest of the process. On 3.12 it gives `VehicleSessionClosed` with the fix and `FrozenInstanceError: cannot assign to field '__traceback__'` without it. On 3.10 both give `VehicleSessionClosed`, so it only has teeth on 3.11 and later.

MatthewDaggitt (approved):

PR #1279, opened 2026-09-20
fix: unbounded dimensions for samplers

ggustavs:
A variable with no finite bound reaches the sampler as ±infinity, so the sampled range was infinite and every starting point `nan`. Starting points are now drawn from a finite region next to the finite bound, or around the origin when both are infinite, whose size is the new `unbounded_search_distance` parameter. Points are still clipped to the real bounds.


vehicle-python/src/vehicle_lang/loss/_pytorch/samplers.py
```diff
@@ -46,10 +50,31 @@ def __init__(
...
+            unbounded_search_distance: How far to search along a dimension the domain leaves
+                unbounded: from the opposite bound, or from the origin when the dimension
+                is unbounded in both directions (default: 10.0)
         """
         self.num_samples = num_samples
         self.num_steps = num_steps
         self.seed = seed
+        self.unbounded_search_distance = unbounded_search_distance
+
+    def _starting_region(
```
MatthewDaggitt: Maybe we should make these as a static(?) method on the parent class `PyTorchSampler`/`TensorFlowSampler` so that every sampler can automatically access them? I don't imagine this is going to change much between samplers?

If so, probably nicer if it takes `unbounded_search_distance` as a parameter.

vehicle-python/src/vehicle_lang/loss/_pytorch/samplers.py
```diff
@@ -37,7 +37,11 @@ class DefaultPyTorchSampler(PyTorchSampler):
     """
 
     def __init__(
-        self, num_samples: int = 10, num_steps: int = 5, seed: int | None = None
+        self,
+        num_samples: int = 10,
+        num_steps: int = 5,
+        seed: int | None = None,
+        unbounded_search_distance: float = 10.0,
```
MatthewDaggitt, line 61: Either need to document this new parameter which is part of the public interface in the CHANGELOG or need to hard-code the value.

MatthewDaggitt (changes requested):
Otherwise looks great!

MatthewDaggitt (approved):

PR #1280, opened 2026-09-20
fix: untypable loss specs as user errors

ggustavs:
`gradientTypeCheck` turned every failure of the gradient subsystem into a `developerError`, so a specification the subsystem cannot type asked the user to report a bug. The subsystem's own message, which names the offending value, is now reported as a user error. Internal errors raised during typing still pass through unchanged.


MatthewDaggitt (commented):
Hmm I don't think this is the right fix... 

The loss type system should _never_ error as it merely tries to over-approximate whether or not a term has gradients. If it does error then there is a bug in the type system that should be fixed rather than reporting that error to the user...

MatthewDaggitt:
I'm going to close this, but feel to reopen if you think I'm missing something @ggustavs!

PR #1281, opened 2026-09-20
fix: bound variables compared against a broadcast tensor

ggustavs:
Removes the prediction that decided whether an assertion could bound a variable, and adds bounds for variables compared against a broadcast tensor.

`isUsableBound` decided at purification time which assertions `findVarBound` would later be able to solve. Nothing enforced that the two agreed, and it has needed strengthening twice. The prediction only existed because `compileLeaf` discarded the term an assertion came from, so an assertion that no quantifier consumed had nothing left to emit and had to reach `developerError`.

- An unconsumed assertion is now returned to the body of the quantifier whose variables it mentions, where those variables are still in scope.
- `findVarBound` validates the bound it has constructed, rejecting one that mentions the tensor being bounded or anything bound inside it. That is checkable; predicting it is not.
- Constraints carry the term they were derived from, so a returned constraint keeps its original form rather than a reconstructed normal form. This needed `findAllBounds` generalised over its leaf type, which it never inspects.
- Variables compared against a broadcast tensor are now bounded, since `const` is adjoint to reduction: `(forall k. a_k <= b) <=> (max_k a_k) <= b`. So `forall i . x ! i <= y ! 0` gives `y ! 0` the lower bound `max 0.0 (reduceMax x)`, symbolically, with no element-wise expansion.

Based on #1283: the new bound is a stack whose elements disagree about gradients, so it needs the fix there to type.

**Golden changes**

- `issue1272-hidden-slices` gains the sharper bound.
- `fourierMotzkin` has its conjuncts reordered. `pointwiseConjunction` is `+`, so the loss is unchanged.
- `issue1272` becomes `(not B) and A` instead of `not (not A or B)`. Under VehicleLoss `not` is negation and `and`/`or` are `max`/`min`, so both are `max A (-B)`.

Routing `VRatConstTensor`/`VRatStackTensor` through `tryUnblock`, as they were before #1165, does not work. `unblockRatTensorValue` returns both unchanged, so `tryAndUnblock` re-enters `compileLinearExpr` on the same value and OOMs on `mnist-robustness`.


MatthewDaggitt (commented):
Oh doh, of course the variables can live in the constant as well! I think this approach is massively overcomplicating everything though. Can we not just run the code in the `mkAtom` to find the bound variables within the existing constant in the existing `isUsableBound` method and eliminate 95% of the new code?

ChangeLog.md
```diff
@@ -16,6 +16,9 @@
 
 * Fixed internal error with assertions with intra-tensor dependencies e.g. `forall x . x ! 0 < x ! 1`.
 
+* Fixed the same internal error for two further shapes of constraint, e.g.
```
MatthewDaggitt: I don't think we need a separate entry for this given there is already an entry above.

vehicle/src/Vehicle/Backend/Loss/PurifyAssertion.hs
```diff
@@ -115,13 +121,37 @@ tryPurifyRatTensorComparison op (TensorComparisonArgs _pDims rDims e1 e2) = do
...
+  -- Occurrences rather than variables, as `x >= const (x ! 0)` mentions one slice twice.
+  residual <- tensorOccurrences =<< constantVariables (constantValue (expression lexp))
+  let occurrences = Map.unionWith (+) solvable residual
+  return $ case Map.lookupMax occurrences of
+    Nothing -> True
+    Just (innermost, occurrencesOfInnermost) ->
+      -- A tensor mentioned only in the constant cannot be solved for.
+      occurrencesOfInnermost == 1 && innermost `Map.member` solvable
+
+-- | The slices mentioned by the term a constant could not evaluate, which `variablesOf` does not
```
MatthewDaggitt: This isn't very grammatical. Maybe something like "The slice variables mentioned in the term that could not be reduced to a linear slice variable coefficients"?

vehicle/tests/golden/issues/issue1272-hidden-slices/DL2Loss.vcl.golden
```diff
@@ -0,0 +1,10 @@
...
+
+@property;
+p : (Tensor Real -> Tensor Real) -> Tensor Real;
+p f = DL2Loss.pointwiseNegation nil (search [3] (const 0.0 [3]) (const 1.0 [3]) (\x -> DL2Loss.pointwiseConjunction nil (DL2Loss.reduceConjunction [3] (foreach k. DL2Loss.pointwiseLessEqualThan nil (x ! 0) (x ! k))) (DL2Loss.pointwiseNegation nil (DL2Loss.pointwiseGreaterEqualThan nil (f x ! 0) 0.5))))
+
+@property;
+q : (Tensor Real -> Tensor Real) -> Tensor Real;
+q f = DL2Loss.pointwiseNegation nil (search [3] (const 0.0 [3]) (const 1.0 [3]) (\x -> search [3] (const 0.0 [3]) (const 1.0 [3]) (\y -> DL2Loss.pointwiseConjunction nil (DL2Loss.reduceConjunction [3] (foreach k. DL2Loss.pointwiseLessEqualThan nil (x ! k) (y ! 0))) (DL2Loss.pointwiseNegation nil (DL2Loss.pointwiseGreaterEqualThan nil (f x ! 0) 0.5)))))
```
MatthewDaggitt: Hmm shouldn't the lower bound for `y[0]` be something equivalent to `max (x ! 0) (x ! 1) (x ! 2) 0`?

MatthewDaggitt (changes requested):
Looks much better, thanks! A few more minor comments?

ChangeLog.md
ggustavs: Done.

vehicle/src/Vehicle/Backend/Loss/PurifyAssertion.hs
ggustavs: Reworded.

vehicle/tests/golden/issues/issue1272-hidden-slices/DL2Loss.vcl.golden
ggustavs: You were right. It needed `isUsableBound` removed first, since it predicted what `findVarBound` could later solve, so the broadcast case would have been a second predicate that had to agree with the first. `y ! 0` now gets `max 0.0 (reduceMax x)`, via the adjunction between `const` and reduction, so no element-wise expansion.

This is now based on #1283, because the new bound is a stack with mixed gradients and needs that fix to type.

PR #1282, opened 2026-09-20
fix: join element gradients in tensor literals

### LINKED ISSUE #1273: Internal error when a comparison relates slices of two separately quantified tensors
When compiling loss for a specification like this:

```vehicle
type In = Tensor Real [2]
type Out = Tensor Real [1]

@network
classifier : In -> Out

@property
p : Bool
p = forall (x : In) . forall (y : In) . (forall i . 0.0 <= x ! i <= 1.0) and (forall i . 0.0 <= y ! i <= 1.0) and x ! 0 >= y ! 1 => classifier x ! 0 >= 0.5
```

the compiler fails with:

```text
Something went wrong internally. Please report the error shown below to `https://github.com/vehicle-lang/vehicle/issues`.
Error: Unfortunately while trying to determining the parts of the program with gradients for export to a loss function, the following error was encountered:
  Error in file 'unknown' at Line 0, Columns 0-0: unable to work out a valid type for the overloaded expression 'min'.
  The possible options explored were:
    1. Tensor (Real withGradients) ds -> Tensor (Real withGradients) ds -> Tensor (Real withGradients) ds
      - rejected: unable to find a consistent type for the overloaded expression 'min'. In particular 'withoutGradients' is not equal to 'withGradients'.
    2. Tensor (Real withoutGradients) ds -> Tensor (Real withGradients) ds -> Tensor (Real withGradients) ds
      - rejected: unable to find a consistent type for the overloaded expression 'min'. In particular 'withoutGradients' is not equal to 'withGradients'.
    3. Tensor (Real withGradients) ds -> Tensor (Real withoutGradients) ds -> Tensor (Real withGradients) ds
      - rejected: unable to find a consistent type for the overloaded expression 'min'. In particular 'withoutGradients' is not equal to 'withGradients'.
    4. Tensor (Real withoutGradients) ds -> Tensor (Real withoutGradients) ds -> Tensor (Real withoutGradients) ds
      - rejected: unable to find a consistent type for the overloaded expression 'min'. In particular 'withGradients' is not equal to 'withoutGradients'.
Please report this as an issue on Github (https://github.com/vehicle-lang/vehicle/issues/)

Stack:
CallStack (from HasCallStack):
  developerError, called at src/Vehicle/Compile/Type/Subsystem.hs:100:17 in vehicle-0.28.0-inplace:Vehicle.Compile.Type.Subsystem
```

However, it does successfully compile the same specification with the two tensors left unrelated:

```vehicle
type In = Tensor Real [2]
type Out = Tensor Real [1]

@network
classifier : In -> Out

@property
p : Bool
p = forall (x : In) . forall (y : In) . (forall i . 0.0 <= x ! i <= 1.0) and (forall i . 0.0 <= y ! i <= 1.0) and y ! 1 >= 0.25 => classifier x ! 0 >= 0.5
```

The same happens with `-l VehicleLoss` as well as `-l DL2Loss`, and is reproducible with both the current PyPI release, and the dev HEAD, from a clean build with `~/.vehicle` removed and every `.vclo` deleted.

ggustavs:
Fixes #1273 . A tensor literal gave every element one shared element type, so `[1.0, x]` with `x` quantified required `withoutGradients = withGradients. stack` is variadic, so the join cannot be a single `MaxGradients`; it is accumulated over the `iterate` that builds the n-ary type, seeded with `withoutGradients`, and the arity it counts over goes in the element type's slot. `evalSeededIterate` is needed because the iteration now has to be applied to that seed.


vehicle/src/Vehicle/Data/Builtin/Loss/Type.hs
```diff
@@ -412,7 +437,13 @@ convertToLossBuiltins decl = do
             AppendList -> sameFunction f
             Iterate -> sameFunction f
             Transpose -> sameFunction f
-            StackTensor -> sameFunction f
+            -- Each element is typed separately, so the shared type slot holds the arity.
+            StackTensor -> case getExpr accessSpine args of
```
MatthewDaggitt, line 441: Firstly, we should only do this for `Stack` operations over the rational type. We don't want to do it for arbitrary stack types. Therefore we would need to add a special `StackRatTensor` function to the `LossBuiltinFunction` class. See deleted comment on Line 213.

vehicle/src/Vehicle/Data/Builtin/Loss/Type.hs
```diff
@@ -412,7 +437,13 @@ convertToLossBuiltins decl = do
...
+              Just stackArgs ->
+                return $
+                  normAppList (Builtin p (StandardBuiltinFunction f)) $
+                    mkExpr accessSpine stackArgs {stackType = arityOf (stackElements stackArgs)}
```
MatthewDaggitt: We shouldn't really be hacking the `StackTensorArgs` class like this. Instead we should just be constructing the arguments manually.

ChangeLog.md
```diff
@@ -16,6 +16,9 @@
 
 * Fixed internal error with assertions with intra-tensor dependencies e.g. `forall x . x ! 0 < x ! 1`.
 
+* Tensor literals now take the join of their elements' gradients, so a literal may mix a constant
```
MatthewDaggitt: This detail isn't relevant to the users. I would say that we fixed an error in the translation when constructing tensors whose elements are quantified variables.

vehicle/src/Vehicle/Compile/Normalise/Builtin.hs
```diff
@@ -723,6 +723,16 @@ evalIterate (IterateArgs t f n e) = do
...
 
+evalSeededIterate ::
+  forall m expr thunk builtin.
+  (MonadLogger m, NormalisableExpr expr thunk builtin m, HasNatExpr expr thunk builtin, BuiltinHasIterate builtin) =>
+  EvalSimple expr thunk SeededIterateArgs builtin m
+evalSeededIterate (SeededIterateArgs args seed) = do
+  result <- evalIterate args
+  case result of
+    Evaluated value | not (null seed) -> Evaluated . exprToThunk <$> forceApp value seed
+    _ -> return result
```
MatthewDaggitt, line 734: I don't quite understand why we needed this notion of `SeededIterate` if all we are doing is applying it to further arguments. If we add an extra argument on it, why doesn't the evaluation algorithm automatically perform the evaluation once it has finished evaluating iterate?

MatthewDaggitt (changes requested):
Thanks for the PR! So this is definitely the right idea but there a few subtleties that this PR is missing.

vehicle/src/Vehicle/Data/Builtin/Loss/Type.hs
ggustavs: Done.

vehicle/src/Vehicle/Data/Builtin/Loss/Type.hs
ggustavs: Done, arguments are built manually now.

ChangeLog.md
ggustavs: Done.

vehicle/src/Vehicle/Compile/Normalise/Builtin.hs
ggustavs: `forceBuiltinEval` matches the whole spine at once, and the `IsArgs` accessors only match an exact length. Over-applying `iterate` makes `getExpr` return `Nothing`, so it is rebuilt unreduced and the extra argument never triggers evaluation.

Could instead make the normaliser handle over-application generically and delete `SeededIterate`. That touches every builtin, so I did not do it here. Prefer that?

PR #1283, opened 2026-09-20
Bugfix/vector literal gradients

ggustavs:
The same fix as #1281 for vecLit, which is what a literal at `Vector Real n` compiles to. A vector's element type is arbitrary, so rather than a shape the rule takes the element type as a family over the gradient, built by abstracting the `Real`s out of the element type the standard program recorded. Literals whose element type contains no `Real` keep the standard typing, since nothing would constrain their element gradients.

Stacked on #1281, so should only be merged after it.


MatthewDaggitt:
Will review this after #1281 is merged and this is updated!

PR #1285, opened 2026-09-21
fix: let bindings in the loss backend

ggustavs:
Fixes two internal errors when a property contains a `let` binding.

The loss backend pushes lambda binders onto the contexts it tracks names and bound variables in, but not `let` binders, so everything under a `let` resolved against the wrong context.

- `Vehicle.Backend.Loss.JSON` did not extend the name context for a `let` body, in either direction of the conversion. A `let` inside a quantifier crashed.
- The domain pass did not extend the bound context for a `let` body. A `let` scoping over a quantifier crashed.

Both now mirror the `Lam` case directly next to them.

Golden tests `loss-letInQuantifier` and `loss-letOutsideQuantifier` cover the two cases. Each fails without its own fix and passes without the other, so they are independent.


MatthewDaggitt (approved):
Wow, this was a really embarrassing bug.

